PR #12, opened 2015-03-02
Student names should be sorted alphabetically by in the teacher competency grid

jmealo:
https://www.wrike.com/open.htm?id=34235442


sencha-workspace/pages/src/page/TeacherCompetencyDashboard.js
```diff
@@ -29,21 +30,24 @@ Ext.define('Site.page.TeacherCompetencyDashboard', {
...
             renderTo: teacherDashboardCt,
             contentArea: contentAreaData
         });
 
-
         // load data embedded in page
-        Ext.getStore('cbl-students-loaded').loadRawData(studentsData);
+        studentsDataStore = Ext.getStore('cbl-students-loaded');
+
+        studentsDataStore.sorters.add(new Ext.util.Sorter({
```
themightychris: You can just pass the config rather than instantiating the Ext.util.Sorter class here, so delete `new Ext.util.Sorter(` and just call `.add({...})`

See http://docs.sencha.com/extjs/5.0/5.0.1-apidocs/#!/api/Ext.util.Collection-cfg-sorters

themightychris: Actually this would be better specified as a `sorters` config within the declaration of the store model

jmealo:
@themightychris issues addressed, good catch.

PR #17, opened 2015-03-05
Full skill language rollover

jmealo:
https://www.wrike.com/open.htm?id=38474811

This PR also includes:
* The latest stylesheets and placeholders for recent activity and graduation progress from @ryon 
* Student dashboard JavaScript files from WIP pull request (all progress)


php-classes/Slate/CBL/CompetenciesRequestHandler.php
```diff
@@ -171,13 +171,21 @@ public static function handleDemonstrationsRequest(Competency $Competency)
...
         // query demonstrations sums
         try {
-            $skillDemonstrations = DB::allRecords(
-                'SELECT Demonstration.ID AS DemonstrationID, Demonstration.StudentID, DemonstrationSkill.SkillID, DemonstrationSkill.Level'
-                .' FROM (SELECT ID FROM `%s` WHERE CompetencyID = %u) AS Skill'
-                .' JOIN `%s` DemonstrationSkill'
-                .'  ON DemonstrationSkill.SkillID = Skill.ID'
-                .' JOIN (SELECT ID, StudentID FROM `%s` WHERE StudentID IN (%s)) Demonstration'
-                .'  ON Demonstration.ID = DemonstrationSkill.DemonstrationID'
+            $skillDemonstrations = DB::allRecords('
```
jmealo: This is how our DBA liked to format SQL, I think it's much more readable, not that the last character of the verb lines up rather than the first. ``AS``, ``ON`` and ``IN`` typically don't wrap to their own line unless it's getting too long.

sencha-workspace/packages/slate-cbl/src/view/student/skill/OverviewWindowController.js
```diff
@@ -0,0 +1,101 @@
...
+                    skillsCombo.setValue(skillStore.findRecord('ID', skillId));
+                    me.syncDemonstrationsTable();
+                });
+
+            } else {
+                me.syncDemonstrationsTable();
+            }
+        }
+
+        Ext.on('resize', function() {
```
jmealo: I think I did this to "fix" the overview window.

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -0,0 +1,181 @@
...
+
+    listeners: {
+        scope: 'this',
+        click: {
+            fn: 'onListClick',
+            element: 'el',
+            delegate: '.cbl-competency-panels, .cbl-competency-panels'
+            /*
+             },
+             mouseover: {
```
jmealo: is this leftover from the non-css full standard text reveal

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -0,0 +1,181 @@
...
+    onListClick: function(ev, t) {
+        var me = this,
+            targetEl;
+
+        if (targetEl = ev.getTarget('.cbl-skill-demo', me.el, true)) {
+            me.fireEvent('democellclick', me, ev, targetEl);
+        }
+    },
+
+    onSkillNameMouseOver: function(ev) {
```
jmealo: Maybe this can go away? Not sure....

sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js
```diff
@@ -0,0 +1,137 @@
...
+    loadSkills: function(student, competency, skillsList) {
+        var me = this,
+            skillsTpl = Ext.XTemplate.getTpl(me.view, 'skillsTpl'),
+            skills, demonstrations, _renderSkills;
+
+        skillsList.removeCls('skills-unloaded').addCls('skills-loading');
+
+        _renderSkills = function() {
+            var demonstrationsLength = demonstrations.length, demonstrationIndex = 0, demonstration, skill;
+            // group skills by student
```
jmealo: This seems out of place, unless the comment is just wrong... but it's working, demonstrations do render

sencha-workspace/packages/slate-cbl/src/view/student/skill/OverviewWindow.js
```diff
@@ -0,0 +1,87 @@
...
+        competency: null,
+        skill: null
+    },
+
+    title: 'Standard Overview',
+    width: 700,
+    minWidth: 700,
+    shadow: 'frame',
+    center: true,
+    fixed: true,
```
jmealo: This is the fixed version. Behavior is a bit more sensible than the teacher one.

site-root/sass/skeleton/modules/_panels.scss
```diff
@@ -0,0 +1,65 @@
+.panel {
```
themightychris: @ryon should this file get merged upstream to skeleton-v2 or skeleton-v1?

ryon: I guess so. It might not be 100% ready to be useful on other sites but it might be a nice thing to have in Skeleton.

sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js
```diff
@@ -0,0 +1,137 @@
...
+    extend: 'Ext.app.ViewController',
+    alias: 'controller.slate-cbl-student-dashboard',
+    requires: [
+        'Slate.cbl.model.Student',
+        'Slate.cbl.model.Competency',
+        'Slate.cbl.view.student.skill.OverviewWindow'
+    ],
+
+    config: {
+        id: 'slate-cbl-teacher-dashboard', // workaround for http://www.sencha.com/forum/showthread.php?290043-5.0.1-destroying-a-view-with-ViewController-attached-disables-listen-..-handlers
```
themightychris: won't be an issue in practice because we won't be loading the teacher dashboard and student dashboard in the same runtime session, but this should be `slate-cbl-student-dashboard`

site-root/sass/skeleton/modules/_panels.scss
jmealo: :skull:

php-classes/Slate/CBL/CompetenciesRequestHandler.php
themightychris: I don't like using multiline strings for embedded code like SQL cause I like to keep indentation consistent within the PHP without filling the SQL with random whitespace. I wouldn't be mad if you used NOWDOCs for this though

themightychris: +1 on the SQL formatting I really like it

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
themightychris: yes this is left over from the non-css reveal, you can delete it

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
themightychris: delete the onSkillNameMouseOver method

sencha-workspace/packages/slate-cbl/src/view/student/skill/OverviewWindowController.js
themightychris: This is not an appropriate place to attach these listeners, as the beforeshow/show events can be fired multiple times in a Component's lifecycle. This would only be appropriate if they were paired with removing the same listeners in a hide event (which might be a good idea for performance if the application made it common to have a lot of this remaining instantiated while hidden).

beforerender/render/afterrender are events that run once in a Component's lifecycle before/after its DOM tree has been built. render/afterrender is typically where you'd want to attach custom listeners to DOM elements within the component.

Since these events are being attached to external, global events though we'll want to be sure that they get removed with this component gets destroyed, so that we're not leaving a bunch of extra listeners running and trying to center a destroyed component.

ExtJS has an easy mechanism for this called "managed listeners", which flips the syntax a bit such that object you're listening to events on is an argument rather than the subject of the function call. So you would do this instead:

```
overviewWindow.mon(Ext.GlobalEvents, 'resize', function() {
    overviewWindow.center();
});

overviewWindow.mon(Ext.getWin(), 'scroll', function() {
    overviewWindow.center();
});
```

This sets up the listener as one managed by `overviewWindow`, which means the listener will automatically be removed when overviewWindow gets destroyed

themightychris: I'm wary though of what behavior this solution causes when the overviewWindow is bigger than the browser viewport, I feel like it would prevent you from scrolling over its surface. We probably need to do something smarter that checks the bounds so its more like the top edge is "glued" to the top of the browser window when you scroll up above it, and scrolling down drags the bottom edge. Or we'd need to put a maximum height on the overviewWindow and enable inner scrolling somewhere.

sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js
themightychris: Good catch, this comment is a stowaway from https://github.com/SlateFoundation/slate-cbl/blob/master/sencha-workspace/packages/slate-cbl/src/view/teacher/DashboardController.js#L289

this comment should read `// group demonstrations by skill`

php-classes/Slate/CBL/CompetenciesRequestHandler.php
jmealo: NOWDOC doesn't support variables. HEREDOC does, but both will throw the indentation outside of the PHP. I think that concatenating the strings is prone to error. I don't think all of the extra whitespace causes much of an issue.

sencha-workspace/packages/slate-cbl/src/view/student/skill/OverviewWindow.js
themightychris: again I'm concerned with what will happen when the overviewWindow is bigger than the browser window, doing this will require an inner scrolling solution.

My initial design principal as I encountered these issues was to avoid using inner scrolling in any detail windows so that users could treat the page as one large canvas. This might make more sense on the teacher end than it does on the student end though. While the window layout in its previous state had less-than-ideal UX, it didn't have the potential to make information inaccessible like this seems to. I would comment this out for now until we can explore a more thorough solution to bigger-that-viewport windows, which i could see either going the direction of modal with inner scrolling or growing with some smarter placement/resize sugar

sencha-workspace/packages/slate-cbl/src/view/student/skill/OverviewWindowController.js
jmealo: Overview window is a fixed size, the app itself won't work at  < 700px. I see what you're saying.

sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js
jmealo: weird, this is already fixed on my server copy.

site-root/sass/skeleton/modules/_panels.scss
themightychris: @jmealo I pushed ryon's changes here upstream to skeleton-v2 and pulled them down to cbl.v1.slate.is. You can delete your override of this file and remove it from the PR. Use the EmergencePullTool to pull the new version down into your instance from cbl.v1.slate.is after you delete the override

sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js
themightychris: No it's not: http://slate2.jeffreymealo.com/develop#/sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js

Maybe you were looking at OverviewWindowController

sencha-workspace/packages/slate-cbl/src/view/student/skill/OverviewWindowController.js
themightychris: Overview window has an initial width of 700px and a minimum width of 700px, but the user can grow it horizontally to be wider. It has no default or minimum/maximum height though, its height is sized to its contents

sencha-workspace/pages/src/page/StudentCompetencyDashboard.js
```diff
@@ -5,180 +5,31 @@ Ext.define('Site.page.StudentCompetencyDashboard', {
     requires: [
         'Site.Common',
         'Slate.cbl.model.Student',
-        'Slate.cbl.model.ContentArea',
-        'Slate.cbl.model.Competency',
+        'Slate.cbl.view.student.Dashboard',
+        'Ext.QuickTips',
         'Ext.XTemplate'
```
themightychris: @jmealo the XTemplate requirement can be removed now that this class isn't using it directly anymore. Looks like the student model isn't be used here anymore either

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -0,0 +1,154 @@
...
+
+    config: {
+        contentArea: null,
+        popover: {
+            pointer: 'none'
+        }
+    },
+
+    competenciesTpl: [
+        '<tpl for="competencies">',
```
themightychris: @jmealo your robot formatter deleted the indentation these strings had for the HTML they contain

sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js
```diff
@@ -0,0 +1,137 @@
...
+            model: 'Slate.cbl.model.Competency'
+        });
+    },
+
+
+    //event handlers
+    onComponentRender: function(dashboardView) {
+        var me = this,
+            siteEnv = window.SiteEnvironment || {},
+            student = siteEnv.cblStudent && Ext.create('Slate.cbl.model.Student', siteEnv.cblStudent)
```
themightychris: @jmealo missing comma, accidental globals ahoy!

sencha-workspace/packages/slate-cbl/src/view/student/DashboardController.js
```diff
@@ -0,0 +1,137 @@
...
+            storeId: 'cbl-competencies-loaded',
+            model: 'Slate.cbl.model.Competency'
+        });
+    },
+
+
+    //event handlers
+    onComponentRender: function(dashboardView) {
+        var me = this,
+            siteEnv = window.SiteEnvironment || {},
```
themightychris: @jmealo I don't want any code in the slate-cbl package referencing window.SiteEnvironment. That's why for TeacherDashboard the Site.page.* class handles all the reading from window.SiteEnvironment and loading into ExtJS stores or passing into the view constructor.

sencha-cbl should be a library that's agnostic to the environment it's loaded into, and window.SiteEnvironment is a detail of its loading environment. the Site.page.* class serves as the "glue" between the .tpl presenter and the reusable components

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -0,0 +1,154 @@
...
+                    blocks.push({ID: skill.ID});
+                }
+
+                return blocks;
+            }
+        }
+    ],
+
+    html: [
+        '<ul class="cbl-competency-panels competencies-unloaded" id="studentDashboardCompetenciesList"></ul>'
```
themightychris: @jmealo javascript components should _never_ create DOM elements with static IDs. Ext will assign them on the fly as needed. I would remove the ID and select this via css class.

However we can simplify this a bit and reduce a level of nesting. Ext.Component has a few techniques for generating initial markup and grabbing references to them. Since we only need a reference to the outermost component the best thing to use here is the [`autoEl`](http://docs.sencha.com/extjs/5.0/5.0.0-apidocs/#!/api/Ext.Component-cfg-autoEl) config, which lets you change the DOM element created at the top level for the Ext.Component from its default of a div to any DomHelper spec you want:

```
autoEl: {
    tag: 'ul',
    cls: 'cbl-competency-panels competencies-unloaded'
}
```

Then `this.el` gets you to the Ext.Element you need and no query is necessary

PR #18, opened 2015-03-08
Added student sorting and m-styling

jmealo:


sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -47,60 +47,95 @@ Ext.define('Slate.cbl.view.student.Dashboard', {
...
         '</tpl>',
-        {
-            getDemonstrationBlocks: function(skill, studentId) {
-                var demonstrationsRequired = skill.DemonstrationsRequired,
-                    blocks = Ext.isArray(skill.demonstrations) ? skill.demonstrations : [];
-
-                blocks = Ext.Array.sort(blocks, function(a, b) {
-                    return a.Level < b.Level ? -1 : 1;
+        
+        {    sortDemonstrations: function sortDemonstrations(demonstrations, limit) {
```
themightychris: shouldn't this opening curly be on its own line?

themightychris: also I don't see the value in breaking sort out into its own function when its used exactly once by getDemonstrationBlocks, that's not likely to change, and getDemonstrationBlocks doesn't do much more than this

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -47,60 +47,95 @@ Ext.define('Slate.cbl.view.student.Dashboard', {
...
+                'use strict';
+                demonstrations = Ext.isArray(demonstrations) ? demonstrations : [];
+                
+                var len = demonstrations.length,
+                    x = 0,
+                    i = 0,
```
themightychris: no need to initialize the iterators here if they'll be initialized at the top of each for block

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -47,60 +47,95 @@ Ext.define('Slate.cbl.view.student.Dashboard', {
...
+                for (x = 0; len > x; x++) {
+                    demo = demonstrations[x];
+                    sortedDemonstrations[demo.Level] = sortedDemonstrations[demo.Level] || [];
+                    sortedDemonstrations[demo.Level].push(demo);
+                }
+                
+                // Sort the levels of demonstrations observed from greatest to least
+                levelsSorted = Object.keys(sortedDemonstrations).sort(function (a, b) {
+                    a = parseInt(a, 10);
+                    b = parseInt(b, 10);
```
themightychris: couldn't we run `parseInt` in the above `for` loop and substantially reduce the # of calls to it?

jmealo: It's only called once per object key. Even though we used a numeric index, ``Object.keys`` will return them as strings. This allows us to do a numeric comparison. This loop should only run 1-3 times typically (or however many levels were logged on a given score).

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
jmealo: I wondered the same thing. Ideally functions should be very small. I want to move ``sortDemonstrations`` to a file that's shared between the student and teacher dashboards. The more modular I make it the better I'm able to re-use it as business logic changes.

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -47,60 +47,113 @@ Ext.define('Slate.cbl.view.student.Dashboard', {
...
+                            // Break out early
+                            break levelLoop;
+                        }
                     }
-
-                    if (b.Level == 0) {
-                        return -1;
+                }
+                
+                // Split up scored and missed demonstrations while keeping order by date (oldest first)
```
jmealo: This could be done in the loop above, but if any of the business logic changes, this is much more manageable and easier to understand I think. This last step is critical.

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -47,60 +47,113 @@ Ext.define('Slate.cbl.view.student.Dashboard', {
...
+                displayDemonstrations.sort(sortByOldest).forEach(function(demo) {
```
jmealo: I originally had this as:

```javascript
displayDemonstrations = displayDemonstrations.sort(sortByOldest);

return displayDemonstrations.filter(// not m).concat(displayDemonstrations.filter(// m only));
```

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
```diff
@@ -47,60 +47,113 @@ Ext.define('Slate.cbl.view.student.Dashboard', {
...
+                var demonstrationsRequired = skill.DemonstrationsRequired,
+                    blocks = this.sortDemonstrations(skill.demonstrations, demonstrationsRequired);
+                    
                 // add empty blocks
                 while (blocks.length < demonstrationsRequired) {
-                    blocks.push({ID: skill.ID});
+                    blocks.push({});
                 }
-
+
```
jmealo: Emergence editor does this a lot.

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
jmealo: this sort will never return 0, I suppose that it can be simplified to:
```javascript
   return parseInt(a, 10) > parseInt(b, 10) ? -1 : 1;
```

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
jmealo: I used your ``for(;`` notation previously. If we can set the values there that's much more readable with this number of variables.

themightychris: once there's more than one `for` loop in the function body and the iterators are being reused i move the initialization to the `for` opening and just declare them at the top of the function, so let's get rid of the assignments here

sencha-workspace/packages/slate-cbl/src/view/student/Dashboard.js
themightychris: @jmealo ah, that makes sense i didn't consider unifying the logic between teacher and student dashboards. A singleton class under sencha-workspace/slate-cbl/src/util would probably be the best place to put such a method

jmealo:
@themightychris can you review/merge this?

php-classes/Slate/CBL/DemonstrationSkill.php
```diff
@@ -22,6 +22,9 @@ class DemonstrationSkill extends \ActiveRecord
             'type' => 'tinyint',
             'unsigned' => true
         ]
+        ,'Demonstrated' => [
```
themightychris: @jmealo is this field supposed to be getting added to this model? Currently `Demonstrated` is a column in the `Demonstration` model table which is referenced by the DemonstrationID column in this table. I thought we solved this by adding Demonstration to the SQL added above in CompetenciesRequestHandler.php and scrapped the idea of duplicating the Demonstrated field from the Demonstration table to the DemonstrationSkill table?

sencha-workspace/packages/slate-cbl/src/util/CBL.js
```diff
@@ -0,0 +1,82 @@
+/*jslint browser: true ,undef: true *//*global Ext*/
+Ext.define('Slate.cbl.util.CBL', {
+    singleton : true,
+    config : {},
```
themightychris: Delete the line `config: {}` this is not needed

sencha-workspace/packages/slate-cbl/src/util/CBL.js
```diff
@@ -0,0 +1,82 @@
...
+    
+    constructor : function(config) {
+        this.initConfig(config);
+    },
```
themightychris: Delete this constructor it is not needed

sencha-workspace/packages/slate-cbl/src/view/student/skill/OverviewWindowController.js
```diff
@@ -79,13 +75,16 @@ Ext.define('Slate.cbl.view.student.skill.OverviewWindowController', {
...
+            demonstrationId = me.view.getDemonstration();
+            
         if (skillId && studentId) {
-            demonstrationsTable.setLoading('Loading demonstrations&hellip;'); // currently not visible due to http://www.sencha.com/forum/showthread.php?290453-5.0.x-loadmask-on-component-inside-window-not-visible
+            // currently not visible due to http://www.sencha.com/forum/showthread.php?290453-5.0.x-loadmask-on-component-inside-window-not-visible
+            demonstrationsTable.setLoading('Loading demonstrations&hellip;');
             Slate.cbl.model.Skill.load(skillId, {
                 callback: function(skill) {
                     skill.getDemonstrationsByStudent(studentId, function(skillDemonstrations) {
+                        demonstrationsTable.selectedDemonstrationId = demonstrationId;
```
themightychris: Why are we stashing this value in `demonstrationsTable.selectedDemonstrationId`? I don't see it getting used anywhere in this PR

jmealo: Please see the comments, I couldn't find the proper way to access this in the template and or pass it without using SiteEnvironment.

sencha-workspace/packages/slate-cbl/src/util/CBL.js
jmealo: This is already done.

jmealo: https://github.com/jmealo/slate-cbl/commit/768c0275343f7db0757f15a47f13da9bd2e08bcd

sencha-workspace/packages/slate-cbl/src/util/CBL.js
jmealo: https://github.com/jmealo/slate-cbl/commit/768c0275343f7db0757f15a47f13da9bd2e08bcd

php-classes/Slate/CBL/DemonstrationSkill.php
jmealo: Exactly, this should not be.

sencha-workspace/packages/slate-cbl/src/view/skill/OverviewBody.js
```diff
@@ -0,0 +1,100 @@
...
+                        '</tr>',
+                    '</tpl>',
+                '</table>',
+                {
+                    escapeAndNewline: function(s) {
+                        s = Ext.util.Format.htmlEncode(s);
+                        return Ext.util.Format.nl2br(s);
+                    },
+                    
+                    getClickedDemonstration: function() {
```
jmealo: @themightychris here

php-classes/Slate/CBL/DemonstrationSkill.php
themightychris: @jmealo so do you want to commit a deletion of this change into this PR or just keep a mental note of it and remove it in a later PR?

jmealo:
Replaced by #21

PR #26, opened 2015-04-03
B21 hotfixes

themightychris:
these hotfixes were applied directly in production


themightychris:
@jmealo here's all your production changes

PR #28, opened 2015-04-10
Fix demonstration date

jmealo:
``$Demonstration->Demonstrated`` this was coming from the database as a string, running it through date made all the dates display as 1969.


themightychris:
Ah, the issue here was that this originally used the `$Demonstration->Created` field which is of type `timestamp`, while `$Demonstration->Demonstrated` is of type `date`. ActiveRecord automatically converts timestamp fields to integers, but does not do this for date columns because they do not represent timestamps and can precede 1970. I wonder if it would make the most sense to change Demonstrated to a timestamp even though we're only offering date input currently? All values will be present, would be better recorded as timezone-free points in time, and other applications integrating with the CBL backend might want to store precise demonstration times

jmealo:
Replaced by #29 as per @themightychris

PR #88, opened 2016-02-07
Release: slate-cbl v0.3.8

themightychris:
- Use default history table names


themightychris:
This PR is being replaced by #96 to target different branches

PR #103, opened 2016-04-20
Store fix

aliyeysides:
Fixes to Chris' comments


sencha-workspace/AggregridExample/app/Application.js
```diff
@@ -6,24 +6,21 @@
...
-    ],
-
     views: [
         'MyAggregrid'
     ],
+    
+    stores: [
+        'AggregridExample.store.Student',
+        'AggregridExample.store.TimePeriod',
+        'AggregridExample.store.Absence'
```
themightychris: @aliyeysides you don't need to use the full name of the stores here

PR #123, opened 2016-05-04
Improve/appcontrollers

aliyeysides:
Converts view controllers to app controllers and consolidates slatetasksteacher components under dashboard/dashboardcontroller


sencha-workspace/SlateDemonstrationsStudent/app/controller/MainController.js
```diff
@@ -4,8 +4,8 @@
  *
  * TODO - Replace this content of this view to suite the needs of your application.
  */
-Ext.define('SlateDemonstrationsStudent.view.main.MainController', {
-    extend: 'Ext.app.ViewController',
+Ext.define('SlateDemonstrationsStudent.controller.MainController', {
```
themightychris: @aliyeysides it's already in the controller namespace you don't need to call it controller again, you won't see `Controller` appended to any other global controllers. Just call this `controller.Main` or `controller.Dashboard`

sencha-workspace/SlateDemonstrationsStudent/app/view/main/Main.js
```diff
@@ -12,12 +12,11 @@ Ext.define('SlateDemonstrationsStudent.view.main.Main', {
     requires: [
         'Ext.window.MessageBox',
 
-        'SlateDemonstrationsStudent.view.main.MainController',
+        'SlateDemonstrationsStudent.controller.MainController',
```
themightychris: @aliyeysides regular controllers don't get required by a view, they get required by the app via the `controllers` config

sencha-workspace/SlateDemonstrationsStudent/app/controller/MainController.js
```diff
@@ -4,8 +4,8 @@
...
+    extend: 'Ext.app.Controller',
```
themightychris: @aliyeysides I find it hard to believe this was the only change necessary

the `alias` isn't needed for non-viewcontroller controllers, remove it.

Also these event handlers won't work like the did as a viewcontroller. They worked with the viewcontroller because the view has `listeners` configured with these handlers identified by string name. You need to convert those into `control` configs on the global controller

sencha-workspace/SlateDemonstrationsTeacher/app/controller/Dashboard.js
```diff
@@ -2,8 +2,8 @@
  * TODO:
  * - move rendering responsibilities to the view?
  */
-Ext.define('SlateDemonstrationsTeacher.view.DashboardController', {
-    extend: 'Ext.app.ViewController',
+Ext.define('SlateDemonstrationsTeacher.controller.Dashboard', {
+    extend: 'Ext.app.Controller',
     alias: 'controller.slate-demonstrations-teacher-dashboard',
```
themightychris: @aliyeysides non-viewcontroller controllers shouldn't have any alias defined as they are singletons. Aliases are a tool for creating instances of a certain class via a registering it to a short name, but singletons don't get instantiated so there's no value to polluting the `controller.*` alias namespace

themightychris:
@aliyeysides the global controllers should be written in the order and with the comments you see in the example here: http://discourse.jarv.us/t/controller-best-practices/83

sencha-workspace/SlateDemonstrationsTeacher/app/controller/Dashboard.js
```diff
@@ -12,17 +12,26 @@ Ext.define('SlateDemonstrationsTeacher.view.DashboardController', {
 
         'Slate.cbl.view.teacher.skill.OverviewWindow',
         'Slate.cbl.view.teacher.skill.OverrideWindow',
-        'Slate.cbl.view.teacher.demonstration.EditWindow'
+        'SlateDemonstrationsTeacher.view.Dashboard',
+        'Slate.cbl.view.teacher.demonstration.EditWindow',
```
themightychris: @aliyeysides global controllers should use their special `views` config to load views which applies smart name resolution and creates a getter for the constructor for you.

```
views: [
    'Dashboard'
]
```

will load `SparkClassroomteacher.view.Dashboard` and then create a getter `this.getDashboardView()` so you can use `this.getDashboardView.create({...})` instead of `Ext.create('SparkClassroomteacher.view.Dashboard', {...})`

For views being loaded from packages instead of within the app you can use `@` to specify the full namespace:

```
views: [
    'skill.OverviewWindow@Slate.cbl.view.teacher'
]
```

will generate the getter `this.getSkillOverviewWindowView()`, it's up to you which parts of the name are before/after the @ based on what makes a nice getter

themightychris:
@aliyeysides all class sources making use of `Ext.Msg` should require `Ext.window.MessageBox`

themightychris:
@aliyeysides you did a bunch of work on classes that weren't being loaded or ever used: 5f1d0f9c4f6788c6e173ea37d9ac39df773a9b1d

themightychris:
@aliyeysides I found some other places where Ext.window.MessageBox needed to be required in 8f3ddaf

PR #128, opened 2016-05-10
Fix/teststime

aliyeysides:
-Add build-all-apps-testing script & adds AggregridExample app to existing build-all-apps script
-Update test README
-Modify dom render test in SlateTasksStudent


themightychris:
@aliyeysides FYI your description in this issue doesn't use valid markdown and misses out on being rendered as a proper bulleted list

Issue #114, opened 2016-04-21
Convert DashboardController ViewControllers to app controllers


themightychris:
There is a teacher.DashboardController and student.DashboardController that extend ViewController and served as the primary controllers for each UI.

Convert both to app controllers in SlateDemonstrationsTeacher and SlateDemonstrationsStudent. This will mostly involve adding the dashboard xtype as a prefix to ref/control ComponentQueries, adding a ref for grabbing the instance of Dashboard already being created by the Viewport controllers, and using that ref instead of ViewController's built-in `this.getView()`

themightychris:
@aliyeysides only one of the two controllers described in this issue have been refactored, `Slate.cbl.view.student.DashboardController` is still around

PR #132, opened 2016-05-21
Remove single-app components from slate-cbl package

themightychris:


themightychris:
@aliyeysides please rebase this branch against the latest `releases/v2/develop` and force-push it to github when it's ready to update this PR

sencha-workspace/SlateDemonstrationsStudent/app/controller/OverviewWindow.js
```diff
@@ -1,22 +1,34 @@
-Ext.define('Slate.cbl.view.student.skill.OverviewWindowController', {
-    extend: 'Ext.app.ViewController',
-    alias: 'controller.slate-cbl-student-skill-overviewwindow',
+Ext.define('SlateDemonstrationsStudent.controller.OverviewWindow', {
```
themightychris, line 1: @aliyeysides per issue #114 you were only supposed to convert the Dashboard controllers from ViewControllers to regular controllers.

Each app is focused on presenting a single dashboard, but many overview windows can be open at a time which IS where we want to use ViewControllers. Your current implementation will fail because the `skillCombo` ref will grab whatever component it finds first and always use the same one.

sencha-workspace/SlateDemonstrationsStudent/app/controller/Viewport.js
```diff
@@ -11,22 +11,25 @@ Ext.define('SlateDemonstrationsStudent.controller.Viewport', {
     
     // controller configuration
     views: [
-        'Dashboard@Slate.cbl.view.student',
-        'RecentProgress@Slate.cbl.view.student'
+        'CompetencyCard',
+        'Dashboard',
+        'DashboardController',
```
themightychris, line 16: @aliyeysides `DashboardController` shouldn't exist anymore, and it definitely shouldn't be getting loaded by an application controller's views list. A ViewController should get loaded by the view that's bound to it (so in this case your controller requires `Dashboard` via views which in-turn requires its controller) -- but in this case it should be gone entirely

sencha-workspace/SlateDemonstrationsStudent/app/controller/Viewport.js
```diff
@@ -11,22 +11,25 @@ Ext.define('SlateDemonstrationsStudent.controller.Viewport', {
...
+        'RecentProgress',
+        'OverviewWindow'
```
themightychris, line 18: @aliyeysides the Dashboard controller is where OverviewWindow instances are getting created, and that's where this line under views should be. The only two components that get caused to be created within this file are Dashboard and RecentProgress (via the `autoCreate: true`+`xtype` configs on their refs.

`CompetencyCard`s could be said to be "caused" to be created across two places:
- the `loadByStudentsAndCompetencies` callback in the Dashboard controller calls `dashboardView.add(...)` for each competency card that needs to be created
- the `defaultType` configured within the Dashboard **view** though is what causes the components added by the controller to be of this type.

So, CompetencyCard should be required by the Dashboard view (which it already is, so can be removed from any controller)

The rule of thumb to follow is that a dependency should be listed (via requires in any class or via shorthand views/stores/models configs in controllers) in every _file_ where the specific class name or xtype is used.

sencha-workspace/SlateDemonstrationsTeacher/app/Application.js
```diff
@@ -13,7 +13,10 @@ Ext.define('SlateDemonstrationsTeacher.Application', {
 
     controllers: [
         'Viewport',
-        'Dashboard'
+        'Dashboard',
+        'EditWindow',
+        'OverviewWindow',
+        'OverrideWindow'
```
themightychris, line 19: @aliyeysides all these windows need to go back to being ViewControllers too

PR #140, opened 2016-06-02
Fix views config loading incorrect OverviewWindow class

aliyeysides:
fixing some issues after I merged slate-package branch into v2/develop


aliyeysides:
@themightychris this is ready to review/merge, has some things I missed during the merging of slate-package bronch

PR #142, opened 2016-06-06
Convert Overviewwindow app controller back to view controller

themightychris:


sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
```diff
@@ -22,7 +22,8 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
    // controller configuration
     views: [
         'SlateDemonstrationsStudent.view.CompetencyCard',
-        'Slate.cbl.view.student.skill.OverviewWindow'
+        'SlateDemonstrationsStudent.view.OverviewWindow',
+        'SlateDemonstrationsStudent.view.OverviewWindowController'
```
themightychris: @aliyeysides you should not be requiring the ViewController directly, the view is what requires and binds to its viewcontroller, your app controller should just be requiring the view. I originally noted this here: https://github.com/SlateFoundation/slate-cbl/pull/132#discussion_r65588891

Also, one of the benefits of using the specialized `views` config in app controllers to require things is that it knows to automatically prepend `AppName.view.` if you just provide it a class name, so this should just be:

Finally, as I noted here this is not the place where CompetencyCard should be getting required: https://github.com/SlateFoundation/slate-cbl/pull/132#discussion-diff-65590087

```
views: [
    'OverviewWindow'
]
```

sencha-workspace/SlateDemonstrationsStudent/app/controller/Viewport.js
```diff
@@ -13,9 +13,9 @@ Ext.define('SlateDemonstrationsStudent.controller.Viewport', {
     views: [
         'CompetencyCard',
         'Dashboard',
-        'DashboardController',
         'RecentProgress',
-        'OverviewWindow'
+        'OverviewWindow',
+        'OverviewWindowController'
```
themightychris: @aliyeysides all the same  things I said above apply here, PLUS you should not be putting the same values in multiple controller's `views` config

sencha-workspace/SlateDemonstrationsTeacher/app/view/EditWindow.js
```diff
@@ -242,14 +274,15 @@ Ext.define('SlateDemonstrationsTeacher.view.EditWindow', {
...
-    
+
     initComponent: function() {
         var me = this,
             studentCombo;
-            
+
         me.callParent(arguments);
-        
-        studentCombo = Ext.ComponentQuery.query('combobox[reference=studentCombo]')[0];
```
themightychris, line 252: @aliyeysides FYI you should _never_ be using Ext.ComponentQuery.query within your code (there are rare legit reasons to, but generally it should only be something you use from your console when debugging)

This code as it was would have grabbed any combobox off the screen (if multiple editwindows were open) or crashed if none were open. It doesn't matter since we're back to using the lookupReference feature of ViewControllers, but what should have been done before FYI was `this.down(`combobox[reference=studentCombo]')` so that the query would be scoped to the EditWindow at hand

Issue #146, opened 2016-06-08
Update Demonstrations Report to include Overrides


ckunkel:
When overrides were created, we did not update the demonstrations export to specify the override. Currently, it uses the DemonstratedLevel (in cbl_demonstration_skills table) to populate the Rating column which is correct for non Overrides. For those demonstrations that were Overridden (Override column in cbl_demonstration_skills table), the Rating should be "O" rather than the DemonstratedLevel.

ckunkel:
http://ferguson.b-21.org/cbl/exports/demonstrations?students=group+class_of_2019

BillClinton:
@ckunkel I believe what we need here is to change line 66 of https://github.com/SlateFoundation/slate-cbl/blob/releases/v2/develop/site-root/cbl/exports/demonstrations.php from

```
$row['Rating'] = $DemonstrationSkill->DemonstratedLevel > 0 ?  $DemonstrationSkill->DemonstratedLevel : 'M';
```

to something like:

```
if ($DemonstrationSkill->Override==1) {
    $row['Rating'] = "O";
} else {
    $row['Rating'] = $DemonstrationSkill->DemonstratedLevel > 0 ?  $DemonstrationSkill->DemonstratedLevel : 'M';
}
```

I need to get a testing environment set up to confirm

PR #151, opened 2016-06-11
Issue146

BillClinton:
These commits address issue #146, "Update Demonstrations Report to include Overrides" and a request to remove the php timeout limit for the export.


themightychris:
@BillClinton check out the tweaks I made before merging

PR #153, opened 2016-06-13
Improve/viewport controllers

aliyeysides:
This branch will remove all viewport controllers in the respective slate-cbl apps and move any functionality away into the Dashboard controllers


sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
```diff
@@ -15,15 +15,66 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
 
    // controller configuration
     views: [
-        'OverviewWindow'
+        'OverviewWindow',
+        'RecentProgress',
+        'CompetencyCard'
```
themightychris: @aliyeysides  The `CompetencyCard` type doesn't get created from this controller, that type gets used in `SlateDemonstrationsStudent.view.Dashboard` where it's already being required.

There are three view types that get created by this controller: `Dashboard`, `RecentProgress`, and `OverviewWindow` (in that same order as they're defined within this very file)

Currently when you load the app this error appears in the console because the `Dashboard` view isn't being required anywhere and ExtJS has to freeze the app, guess where the view might be, and synchronously load it. Luckily it works in this case, but warnings like these are not to be ignored and added by PRs

screen shot 2016-06-15 at 9 55 56 am

sencha-workspace/SlateDemonstrationsTeacher/app/controller/Dashboard.js
```diff
@@ -47,10 +47,41 @@ Ext.define('SlateDemonstrationsTeacher.controller.Dashboard', {
     ],
 
     refs: {
-        dashboardCt: 'slate-demonstrations-teacher-dashboard',
-
         studentProgressGrid: 'slate-demonstrations-teacher-dashboard slate-demonstrations-teacher-studentsprogressgrid',
-        teacherOverviewwindow: 'slate-demonstrations-teacher-skill-overviewwindow'
+        teacherOverviewwindow: 'slate-demonstrations-teacher-skill-overviewwindow',
```
themightychris: @aliyeysides you should not define refs only for sake of using the shorter string in the `control` config. Only define a ref when you need to access the instance from one of the controllers methods -- then you may replace any identical selector under `control` with the ref name.

Further, it is not valid to use a ref for OverviewWindow because OverviewWindow is not a singleton in the context of the application, there can be any-number of windows open at a given time within the application. A string selector in `control` will catch events from any matching instance whereas a ref would cache an arbitrary one of the set of windows.

sencha-workspace/SlateDemonstrationsTeacher/app/controller/Dashboard.js
```diff
@@ -47,10 +47,41 @@ Ext.define('SlateDemonstrationsTeacher.controller.Dashboard', {
...
+        dashboardCt: {
+            selector: 'slate-demonstrations-teacher-dashboard',
+            autoCreate: true,
+
+            xtype: 'slate-demonstrations-teacher-dashboard'
+        }
```
themightychris: @aliyeysides refs should be grouped where there is hierarchy. As you can see in the selectors, the `studentsProgressGrid` is a subcomponent of `dashboardCt` that gets created by it so we should group/order them as seen in the commit i'll make

sencha-workspace/SlateDemonstrationsTeacher/app/controller/Viewport.js
```diff
@@ -1,45 +0,0 @@
-Ext.define('SlateDemonstrationsTeacher.controller.Viewport', {
-    extend: 'Ext.app.Controller',
-
-
-    // controller configuration
-    views: [
-        'Dashboard'
```
themightychris, line 7: @aliyeysides `Dashboard` view getting required went missing in this app too and was generating a warning in the console

sencha-workspace/SlateDemonstrationsTeacher/app/controller/Viewport.js
```diff
@@ -1,45 +0,0 @@
...
-
-
-    // controller templates method overrides
-    onLaunch: function () {
-        var siteEnv = window.SiteEnvironment || {},
-            contentAreaCode = (siteEnv.cblContentArea || {}).Code,
-            dashboardCt, progressGrid;
-
-        dashboardCt = this.getDashboardCt();
-        progressGrid = dashboardCt.getProgressGrid();
```
themightychris, line 27: @aliyeysides these two variable initializations must have been broken out from the declaration block in the past because there was a condition in between that could abort it. In its current form though these should be consolidated

sencha-workspace/SlateTasksTeacher/app/controller/Viewport.js
```diff
@@ -1,34 +0,0 @@
...
-    ],
-    
-    
-    config: {
-    },
-    
-    
-    // controller configuration
-    views: [
-        'Dashboard'
```
themightychris, line 14: @aliyeysides Dashboard view require again going missing and generating a warning

themightychris:
@aliyeysides the Dashboard view in the `SlateTasksTeacher` app was requiring the Dashboard app controller too, creating a circular dependency once the controller was requiring the view as it should. This was probably a holdover from when the view required its viewcontroller

PR #176, opened 2016-07-20
Issues/167

BillClinton:
Makes various elements in the Student task details dynamic, including attachments, comments, and the rating widget.  Also includes change to the way status messages and css classes are determined in the task tree.


BillClinton:
@ChrisHoopes This can be merged

PR #218, opened 2016-07-25
Use natural heights for student task expand/collapse

ryon:
Not sure my solution is optimal (I ran into race conditions with the DOM rendering, and the `height` transition seems sluggish with a lot of tasks), but it should work for now.
#217

### DIFF
--- a/sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
+++ b/sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
@@ -216,16 +216,33 @@ Ext.define('SlateTasksStudent.view.TaskTree', {
         }
     },
 
+    afterTasksLoad: function() {
+        var tree = this,
+            sublists = tree.el.select('.slate-tasktree-sublist');
+
+        sublists.each(function(sublist) {
+            sublist.set({
+                'data-natural-height': sublist.getHeight(),
+                style: {
+                    height: 0
+                }
+            });
+        });
+    },
+
     onTreeClick: function(ev, t) {
         var target = Ext.get(t),
-            parentEl,recordId;
+            treeItem = target.up('.slate-tasktree-item'),
+            sublist = treeItem.down('.slate-tasktree-sublist'),
+            recordId;
 
-        if (target.is('.slate-tasktree-nub.is-clickable')) {
-            target.up('.slate-tasktree-item').toggleCls('is-expanded');
+        if (target.is('.slate-tasktree-nub.is-clickable') && treeItem && sublist) {
+            var newHeight = treeItem.hasCls('is-expanded') ? 0 : sublist.getAttribute('data-natural-height');
+            sublist.setHeight(newHeight);
+            treeItem.toggleCls('is-expanded');
         } else {
-            parentEl = target.up('.slate-tasktree-item');
-            if (parentEl) {
-                recordId = parentEl.dom.getAttribute('recordId');
+            if (treeItem) {
+                recordId = treeItem.dom.getAttribute('recordId');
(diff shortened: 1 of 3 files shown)


sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
```diff
@@ -216,16 +216,33 @@ Ext.define('SlateTasksStudent.view.TaskTree', {
...
     },
 
+    afterTasksLoad: function() {
+        var tree = this,
+            sublists = tree.el.select('.slate-tasktree-sublist');
+
+        sublists.each(function(sublist) {
+            sublist.set({
+                'data-natural-height': sublist.getHeight(),
+                style: {
```
themightychris, line 226: @ryon you should use the `.setStyle()` api for this or the convenient wrapper `.setHeight()`, I'm concerned this form may be liable to overwrite other style attributes.

sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
```diff
@@ -216,16 +216,33 @@ Ext.define('SlateTasksStudent.view.TaskTree', {
         }
     },
 
+    afterTasksLoad: function() {
+        var tree = this,
+            sublists = tree.el.select('.slate-tasktree-sublist');
+
+        sublists.each(function(sublist) {
```
themightychris, line 223: @ryon to avoid any race conditions, update the top of this function like this:

``` javascript
var tree = this;

if (!tree.rendered) {
    tree.on('render', 'afterTasksLoad', this, { single: true} });
    return;
}

tree.el.select('.slate-tasktree-sublist').each(function(sublistEl) {
    // ...
});
```

sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
```diff
@@ -216,16 +216,33 @@ Ext.define('SlateTasksStudent.view.TaskTree', {
...
-            parentEl,recordId;
-
-        if (target.is('.slate-tasktree-nub.is-clickable')) {
-            target.up('.slate-tasktree-item').toggleCls('is-expanded');
+            treeItem = target.up('.slate-tasktree-item'),
+            sublist = treeItem.down('.slate-tasktree-sublist'),
+            recordId;
+
+        if (target.is('.slate-tasktree-nub.is-clickable') && treeItem && sublist) {
+            var newHeight = treeItem.hasCls('is-expanded') ? 0 : sublist.getAttribute('data-natural-height');
```
themightychris, line 240: @ryon don't use the `var` keyword within non-function blocks -- they don't create a variable scope. The `newHeight` variable should be declared above and only set here, like recordId

sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
```diff
@@ -216,16 +216,33 @@ Ext.define('SlateTasksStudent.view.TaskTree', {
...
+            sublist.setHeight(newHeight);
+            treeItem.toggleCls('is-expanded');
         } else {
-            parentEl = target.up('.slate-tasktree-item');
-            if (parentEl) {
-                recordId = parentEl.dom.getAttribute('recordId');
+            if (treeItem) {
+                recordId = treeItem.dom.getAttribute('recordId');
```
themightychris, line 245: @ryon recordId should get turn into an integer before being passed to a component event with `parseInt(..., 10)`

themightychris:
@ryon this approach looks generally optimal to me, given the current external constraints. Please review and assign this PR back to me when it's ready to review/merge

ryon:
Outdated now, replaced by #290

PR #234, opened 2016-08-01
Configure teacher tasks manager

nbey:
TO DO:
- Configure Grid
  - [x] Implement Aggregrid
- Configure User Actions
  - [x] Create Task
  - [x] Rate Student Task
  - [x] Assign Student Task
  - [x] Edit Student Task
  - [x] Unassign Student Task

### DIFF
--- a/sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
+++ b/sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
@@ -207,6 +207,8 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
                 return me.editStudentTask(studentTask);
             } else if (studentTask.get('TaskStatus') === 'submitted' || studentTask.get('TaskStatus') === 're-submitted') {
                 return me.rateStudentTask(studentTask);
+            } else if (studentTask.get('TaskStatus') === 'completed') {
+                return me.rateStudentTask(studentTask, true);
             }
         }
 
@@ -412,7 +414,7 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
         taskEditor.show();
     },
 
-    rateStudentTask: function(studentTask) {
+    rateStudentTask: function(studentTask, readOnly) {
         var me = this,
             taskRater = me.getTaskRater(),
             task = me.getTasksStore().getById(studentTask.get('TaskID'));
@@ -426,6 +428,7 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
 
         taskRater.setTask(task);
         taskRater.setStudentTask(studentTask);
+        taskRater.setReadOnly(readOnly);
 
         taskRater.show();
     },
--- a/sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
+++ b/sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
@@ -31,7 +31,8 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
     ],
     models: [
         'Task@Slate.cbl.model',
-        'StudentTask@Slate.cbl.model'
+        'StudentTask@Slate.cbl.model',
+        'Comment@Slate.cbl.model.tasks'
     ],
 
 
@@ -68,6 +69,7 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
         },
         taskEditorForm: 'slate-tasks-teacher-taskeditor slate-modalform',
         skillsField: 'slate-tasks-teacher-taskeditor slate-skillsfield',
+        commentsField: 'slate-tasks-teacher-taskrater slate-commentsfield',
         attachmentsField: 'slate-tasks-teacher-taskeditor slate-tasks-attachmentsfield',
         assignmentsField: 'slate-tasks-teacher-taskeditor slate-tasks-assignmentsfield',
         assignmentsComboField: 'slate-tasks-teacher-taskeditor slate-tasks-assignmentsfield combo',
@@ -98,6 +100,9 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
         },
         assignmentsComboField: {
             render: 'onAssigneeComboRender'
+        },
+        commentsField: {
+            publish: 'onCommentsFieldPublish'
         },
 
         'slate-tasks-teacher-taskrater button[action=accept]': {
@@ -196,6 +201,28 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
         comboStore.removeAll();
         comboStore.add(studentsStore.getRange());
         combo.setValueOnData();
+    },
+
+    onCommentsFieldPublish: function(fieldContainer, field) {
+        var me = this,
+            record = fieldContainer.getRecord(),
+            comment,
+            originalComments = record.get('Comments') || [];
+
+        comment = me.getCommentModel().create({
+            Message: field.getValue(),
+            ContextID: record.getId(),
+            ContextClass: record.get('Class')
+        });
+
+        comment.save({
+            success: function(rec) {
+                originalComments.push(rec.getData({ serialize: true }));
+                record.set('Comments', originalComments);
+                fieldContainer.updateRecord(record);
+                field.setValue('');
+            }
+        });
     },
 
     onAcceptTaskClick: function() {
--- a/sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
+++ b/sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
@@ -1,7 +1,12 @@
-/* global Slate */
+/* jslint browser: true, undef: true, laxcomma:true *//* global Ext*/
 /**
- * TODO:
- * - move rendering responsibilities to the view?
+ * The Dashboard controller manages the main functionality of the SlateTasksTeacher application where teachers can
+ * browse, search, create, edit, and assign tasks.
+ *
+ * ## Responsibilities
+ * - Handle section/:sectionId route
+ * - Handle CRUD operations for tasks/student tasks
+ * - Filter StudentsGrid tasks/students by selected section
  */
 Ext.define('SlateTasksTeacher.controller.Dashboard', {
     extend: 'Ext.app.Controller',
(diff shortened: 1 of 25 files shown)


sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -1,43 +1,457 @@
...
+        },
+        'slate-tasks-teacher-appheader button[action=create]': {
+            click: 'onCreateTaskClick'
+        }
+    },
+
+    routes: {
+        'section/:sectionId': {
+            sectionId: '([a-zA-Z0-9])+',
+            action: 'viewCourseSection'
```
themightychris: @nbey our convention for naming route handles in `showCourseSection(sectionID)`. Please see https://github.com/JarvusInnovations/jarvus-guide-to-extjs/blob/master/applications/controllers.md

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -1,43 +1,457 @@
...
+            datepicker.show().showAt(coords[0], coords[1] - datepicker.getHeight()); // subtract datepicker height from y coord to show above button.
+        }
+    },
+
+    onUnassignStudentTaskClick: function() {
+        var me = this,
+            taskRater = me.getTaskRater(),
+            studentTask = taskRater.getStudentTask();
+
+        Ext.Msg.confirm('Are you sure?', 'Do you want to unassign this student task? This can not be undone.', function(ans) {
```
themightychris, line 284: @nbey `Ext.MessageBox` needs to be explicitly listed in this class's `requires` if you're going to use the `Ext.Msg` singleton it provides

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -1,43 +1,457 @@
...
+        var me = this,
+            coords = btn.getXY(),
+            datepicker;
+
+        if (btn.dateSelected) {
+            me.assignStudentTaskRevision(btn.dateSelected);
+        } else {
+            datepicker = Ext.widget({
+                xtype: 'datepicker',
+                floating: true,
```
themightychris, line 268: @nbey this controller needs to explicitly depend on the datepicker component via its `requires` array.

Also, you can use the ref system to cleanly define a "factory" for these pickers:

``` javascript
refs: {
    datePicker: {
        forceCreate: true,

        xtype: 'datepicker',
        floating: true
    }
}
```

(I'm not sure if a `selector` is still required, it would be useless with `forceCreate` but give the framework what it wants if it complains about it being missing)

and then you can do:

```
me.getDatePicker({
    handler: function(picker, date) {
        // ...
    }
}).showAt(...);
```

themightychris, line 268: @nbey Also, don't call show() and showBy separately, showBy handles showing and does it with some finer control to ensure it doesn't flash in the wrong position first.

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -1,43 +1,457 @@
...
+
+        if (studentTask) {
+            taskStatus = studentTask.get('TaskStatus');
+            if (taskStatus === 'assigned') {
+                return me.editStudentTask(studentTask);
+            } else if (taskStatus === 'submitted' || taskStatus === 're-submitted' || taskStatus === 're-assigned') {
+                return me.rateStudentTask(studentTask);
+            } else if (taskStatus === 'completed') {
+                return me.rateStudentTask(studentTask, true);
+            }
```
themightychris: @nbey don't use `else` following `if` blocks that end in `return`, just do:

```
if (...) {
    return ...;
}

if (...) {
    return ...;
}
```

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -1,43 +1,457 @@
...
+        return me.assignStudentTask(taskId, studentId);
+    },
+
+    onRateSkillClick: function(ratingView, ratingObject) {
+        var me = this,
+            taskRater = me.getTaskRater(),
+            studentTask = taskRater.getStudentTask();
+
+        Slate.API.request({
+            url: '/cbl/student-tasks/'+studentTask.getId()+'/rate',
```
themightychris: @nbey the `Ext.data.Model` class has a standard `toUrl()` method you can override and implement. Centralize this logic on the model so the code here can just be: `url: studentTask.toUrl() + '/rate'`

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -1,43 +1,457 @@
...
+        taskEditor.setStudentTask(studentTask);
+        taskEditor.show();
+    },
+
+    editTask: function(taskRecord) {
+        var me = this,
+            taskEditor = me.getTaskEditor();
+
+        if (!taskRecord) {
+            taskRecord = Ext.create('Slate.cbl.model.Task');
```
themightychris: @nbey this is an implicit dependency on the model, avoid calling `Ext.create` or `Ext.widget` within a controller.

Put at the top of the controller:

```
models: [
    'Task@Slate.cbl.model'
]
```

and then you can do: `me.getTaskModel().create({ ...})`

sencha-workspace/SlateTasksTeacher/app/store/StudentTasks.js
```diff
@@ -0,0 +1,29 @@
+Ext.define('SlateTasksTeacher.store.StudentTasks', {
+    extend: 'Ext.data.Store',
+    requires: [
+        'Slate.cbl.model.StudentTask'
```
themightychris: @nbey this is redundant, stores add their `model` to the requires list automatically.

sencha-workspace/SlateTasksTeacher/app/store/StudentTasks.js
```diff
@@ -0,0 +1,29 @@
...
+    ],
+
+    model: 'Slate.cbl.model.StudentTask',
+    remoteFilter: true,
+    remoteSort: true,
+    pageSize: 0,
+
+    proxy: {
+        type: 'slate-records',
```
themightychris, line 13: @nbey this class needs to require the proxy class too if it's going to require this `type`. The model could get reconfigured one day to use a different proxy

sencha-workspace/SlateTasksTeacher/app/store/Students.js
```diff
@@ -0,0 +1,21 @@
+Ext.define('SlateTasksTeacher.store.Students',{
+    extend: 'Slate.cbl.store.Students',
+    requires: [
+        'Emergence.ext.proxy.Records'
+    ],
+    proxy: {
```
themightychris: @nbey the proxy config should be down below `config`, not up with the `Ext.define` options

sencha-workspace/SlateTasksTeacher/app/view/AppHeader.js
```diff
@@ -0,0 +1,50 @@
...
+        align: 'left'
+    },
+
+    items: [{
+        xtype: 'component',
+        cls: 'slate-appheader-title',
+        itemId: 'title',
+        html: 'Teacher Task Manager'
+    },{
+        xtype: 'combo',
```
themightychris, line 21: @nbey combo needs to be required

sencha-workspace/SlateTasksTeacher/app/view/Dashboard.js
```diff
@@ -23,6 +28,10 @@ Ext.define('SlateTasksTeacher.view.Dashboard', {
         return Ext.factory(gridLegend, 'SlateTasksTeacher.view.GridLegend', oldGridLegend);
     },
 
+    updateCourseSection: function(courseSection) {
+        this.fireEvent('coursesectionselect', courseSection);
```
themightychris: @nbey the first parameter to any component event should be a reference to the component firing the event (`this` in this case)

sencha-workspace/SlateTasksTeacher/app/view/StudentsGrid.js
```diff
@@ -2,312 +2,233 @@
...
+                    }
+
+                    return html;
+                }
             }
-        ]
-    },
+        ],
+
+        tpl: [
```
themightychris, line 95: @nbey why are all these structural tpls' being overridden in this implementation? They're not supposed to be customized and overriding them here is dangerous

sencha-workspace/SlateTasksTeacher/app/view/StudentsGrid.js
```diff
@@ -2,312 +2,233 @@
...
-        scope: 'this',
-        click: {
-            fn: 'onGridClick',
-            element: 'el',
-            delegate: '.slate-studentsgrid-row'
-        }
+        me.callParent(arguments);
+        me.setConfig({
+            cellRenderer: me.cellRenderFn,
+            subCellRenderer: me.cellRenderFn
```
themightychris, line 214: @nbey why is this necessary? The function bodies should just be declared directly in the config section

sencha-workspace/SlateTasksTeacher/app/view/TaskEditor.js
```diff
@@ -0,0 +1,99 @@
...
+            assignmentsfieldCheckbox = assignmentsfield.down('checkboxfield'),
+            titleField = me.down('slate-tasks-titlefield'),
+            parentTaskField = titleField.next('slate-tasks-titlefield'),
+            experienceField = me.down('#experience-type'),
+            duedateField = me.down('#duedate'),
+            expirationdateField = me.down('#expirationdate'),
+            skillsField = me.down('slate-skillsfield'),
+            attachmentsfield = me.down('slate-tasks-attachmentsfield'),
+            instructionsField = me.down('#instructions'),
+            taskPrivacyField = me.down('#status');
```
themightychris, line 23: @nbey consider providing a `getForm()` method on the modal, grabbing that once, and then using form`s`getField` api for this to grab form field based on their name as intended rather than with itemIds

sencha-workspace/packages/slate-cbl/src/model/StudentTask.js
```diff
@@ -0,0 +1,116 @@
...
+    proxy: {
+        type: 'slate-records',
+        url: '/cbl/student-tasks'
+    },
+
+    getTaskSkillsGroupedByCompetency: function() {
+        var comps = [], compIds = [],
+            skills = this.get('TaskSkills');
+
+        Ext.each(skills, function(skill) {
```
themightychris: @nbey use for loops instead of `Ext.each`

sencha-workspace/packages/slate-cbl/src/view/AttachmentsList.js
```diff
@@ -13,20 +13,29 @@ Ext.define('Slate.cbl.view.AttachmentsList', {
...
+                '<li class="slate-attachmentslist-item <tpl if="kind">slate-attachment-{kind}</tpl>">',
+                    '<span class="slate-attachment-title"><a href="{URL}" target=_blank>{title}</a></span>',
+                    '<tpl if="this.owner.getEditable()">',
+                        '<button class="plain" action="settings"><i class="fa fa-gear"></i></button>',
+                        '<button class="plain" action="remove"><i class="fa fa-times-circle"></i></button>',
+                    '</tpl>',
+                '</li>',
+            '</tpl>'
+        ];
+        me.callParent(arguments);
```
themightychris: @nbey you shouldn't be calling `initConfig`, that already gets called in the component's lifecycle. Why was it necessary to set tpl from initComponent instead of just overriding the property? An initComponent override looking like this shouldn't exist

sencha-workspace/packages/slate-cbl/src/widget/AttachmentsField.js
```diff
@@ -42,6 +46,29 @@ Ext.define('Slate.cbl.widget.AttachmentsField', {
         disabled: true
     }],
 
+    updateReadOnly: function(readOnly) {
+        if (!this.rendered) {
+            return this.on('render', function() {
+                return this.updateReadOnly(readOnly);
+            });
```
themightychris: @nbey be sure to pass the `{ single: true }` option with this `.on()` call, or this component being rendered within something like a tab panel or window that can be rendered/unrendered several times will result in an infinitely growing stack of this function getting called for each update

sencha-workspace/packages/slate-cbl/src/widget/RatingView.js
```diff
@@ -74,5 +45,102 @@ Ext.define('Slate.cbl.widget.RatingView', {
...
+            }];
+        if (!me.getReadOnly()) {
+            if (!menu && menuRatings.length) {
+                Ext.each(menuRatings, function(mr) {
+                    items.push({
+                        text: mr,
+                        value: mr
+                    });
+                });
+                me.setMenu(menu = Ext.create('Ext.menu.Menu', {
```
themightychris, line 127: @nbey this class should explicitly require `Ext.menu.Menu`

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
nbey, line 268: @themightychris I'm going to remove this code during my next work session (we are no longer going to use the datepicker like this)

sencha-workspace/SlateTasksTeacher/app/view/StudentsGrid.js
nbey, line 95: @themightychris I changed the name of the field used in the template, `{id}`, to match the models id field, `{ID}`. Should I create a non-persisted field instead?

sencha-workspace/SlateTasksTeacher/app/view/StudentsGrid.js
nbey, line 214: @themightychris Not necessary, just thought it would be better for them to use the same exact render function without duplicating the code.

sencha-workspace/packages/slate-cbl/src/view/AttachmentsList.js
nbey: @themightychris I was having trouble with the configuring the template -- It seems that the template object doesn't have an updated reference to the container it's displayed in.
`this.owner.getEditable()` appears to _only_ return it's **initial value**. Any suggestions?

sencha-workspace/SlateTasksTeacher/app/view/StudentsGrid.js
themightychris: @nbey better approach (doesn't mess with initComponent lifecycle): 4170cf3f80a7ff81b87717f468b191d6bd6073ac

sencha-workspace/SlateTasksTeacher/app/view/StudentsGrid.js
themightychris: The aggregrid should be adoptable to any store you want to provide, I just resolved the need for this in aggregrid's master branch

PR #280, opened 2016-08-20
API changes

nbey:


php-classes/Slate/CBL/Skill.php
```diff
@@ -111,4 +122,24 @@ public function save($deep = true)
...
+    
+    public function getCompetencyCode()
+    {
+        return $this->Competency ? $this->Competency->Code : null;
+    }
+    
+    public function getCompetencyLevel()
+    {
+        $level = null;
+        if ($GLOBALS['Session']->PersonID && $this->Competency && $StudentCompetency = StudentCompetency::getByWhere(['StudentID' => $GLOBALS['Session']->PersonID, 'CompetencyID' => $this->Competency->ID])) {
```
themightychris, line 139: @nbey this is going to have really poor performance characteristics at scale. I've learned that instead of leaning on dynamicfields too much -- which are easy to implement but super naive because there's zero coordination between populating them for every record in a result -- it's often better to create a UI-centric endpoint that can use optimized joins to deliver a specialized result set efficiently.

You can see this pattern with the demonstrations dashboards, they use the standard CRUD endpoints as much as possible for loading their auxiliary data, but for their primary datasets that are a novel combination of data from across tables there are a couple endpoints within the `Slate\CBL\Demonstartions\*DashboardRequestHandler` classes. 

I setup all the endpoints for the /cbl/dashboards/tasks/\* apps as simple inline PHP scripts that just output the template, but at some point they should get _DashboardRequestHandler classes of their own like the /cbl/dashboards/demonstrations/_ routes do that handle both loading the app and providing for a couple specialized data endpoints

php-classes/Slate/CBL/Tasks/StudentTask.php
```diff
@@ -0,0 +1,162 @@
...
+            $skillRating = StudentTaskSkillRating::getByWhere(['StudentTaskID' => $this->ID, 'SkillID' => $skill->ID]);
+            $taskSkills[] = array_merge($skill->getData(), [
+                'CompetencyLevel' => $studentCompetency ? $studentCompetency->Level : null,
+                'CompetencyCode' => $skill->Competency ? $skill->Competency->Code : null,
+                'CompetencyDescriptor' => $skill->Competency ? $skill->Competency->Descriptor : null,
+                'Rating' => $skillRating ? $skillRating->Score : null
+            ]);
+        }
+        
+        return $taskSkills;
```
themightychris, line 147: @nbey if you're putting together a deep custom data structure that's not just a calculation based on data within each record or a complex relationship, it's probably best to make a custom UI-centric data endpoint than a dynamic field

php-classes/Slate/CBL/Tasks/Task.php
```diff
@@ -2,11 +2,16 @@
...
+use HandleBehavior;
+use Slate\CBL\Skill;
+use Slate\CBL\Tasks\Attachments\AbstractTaskAttachment;
+
 class Task extends \VersionedRecord
 {
     //VersionedRecord configuration
     public static $historyTable = 'history_cbl_tasks';
-
+
```
themightychris, line 14: @nbey watch the trailing whitespace, these are my "user settings" for VSCode that includes an option that prevents any trailing space from being saved:

``` json
// Place your settings in this file to overwrite the default settings
{
    "editor.insertSpaces": true,
    "editor.renderWhitespace": true,
    "explorer.workingFiles.maxVisible": 20,
    "files.trimTrailingWhitespace": true
}
```

php-config/Slate/CBL/Tasks/Task.config.d/search-conditions.php
```diff
@@ -0,0 +1,44 @@
...
+            'className' => Person::class,
+            'localField' => 'CreatorID',
+            'foreignField' => 'ID',
+            'aliasName' => $personTableAlias
+        ],
+        'sql' => 'CONCAT('.$personTableAlias.'.FirstName, " ", '.$personTableAlias.'.LastName) LIKE "%%%s%%"'
+    ]
+];
+
+Task::$searchConditions = array_merge(Task::$searchConditions, $searchConditions);
```
themightychris, line 44: @nbey php-config files shouldn't be used to provide part of the base implementation of a class. If you really need to set some config dynamically, you can do that by overriding the magic __classLoaded function emergence implements for classes (it gets run after config has been loaded but before the autoloader returns control to the first place using the class)

Also, I believe you can set the `'sql'` option for searchConditions to a callable like `[__CLASS__, 'staticMethodName']` and not have to set any config dynamically. Shouldn't the aliasName be getting calculated automatically the same way already?

PR #285, opened 2016-08-21
Fix double loading issue caused by routing changes

BillClinton:


BillClinton:
@hoopes this is ready for Q&A.  This was a behind the scenes fix and there should be no visible difference in routing if working properly

PR #302, opened 2016-08-31
Fixes to student dashboard

BillClinton:


BillClinton:
@ChrisHoopes This can be merged into dev if/when Q&A demonstrates that nothing is broken

PR #300, opened 2016-08-27
Student tasks dashboard fixes and optimization

BillClinton:
work in progress

PR #326, opened 2016-10-08
CBL-140 Optimize performance in student task dashboard

BillClinton:
work in progress


themightychris:
Everything in here is also in #300

PR #334, opened 2016-10-21
Export scripts

BillClinton:


BillClinton:
This is ready for merge.  These are all new files and should present no conflicts.

PR #359, opened 2016-12-21
Release: slate-cbl v2.1.11

nbey:
- Fix issue with showing assigned tasks in SlateTasksTeacher app
- Rename `CourseSectionID` column to `SectionID`
- Fix issue with reopening Submit Evidence window in `SlateDemonstrationsTeacher` app
- Require student tasks to be assigned to students
- Show student task validation errors on failed saves

### DIFF
--- a/php-classes/Slate/CBL/Tasks/StudentTask.php
+++ b/php-classes/Slate/CBL/Tasks/StudentTask.php
@@ -128,7 +128,7 @@ class StudentTask extends \VersionedRecord
         ],
 
         'Student' => [
-            'validator' => 'require-relationship'
+            'method' => [__CLASS__, 'validateStudent']
         ]
     ];
 
@@ -153,6 +153,18 @@ class StudentTask extends \VersionedRecord
 
             default:
                 return parent::getValue($name);
+        }
+    }
+
+    protected static function validateStudent(\RecordValidator $RecordValidator, StudentTask $StudentTask, $options, $validator)
+    {
+        if (!$RecordValidator->hasErrors('Student')) {
+            $Student = $StudentTask->Student;
+            if (!$Student) {
+                $RecordValidator->addError('Student', 'Tasks must be assigned to a student.');
+            } elseif (!$Student instanceof \Slate\People\Student) {
+                $RecordValidator->addError('Student', 'Tasks can only be assigned to students.');
+            }
         }
     }
 
--- a/php-classes/Slate/CBL/Tasks/StudentTask.php
+++ b/php-classes/Slate/CBL/Tasks/StudentTask.php
@@ -185,7 +185,7 @@ class StudentTask extends \VersionedRecord
 
         if ($this->Task && $this->Task->Skills) {
             foreach ($this->Task->Skills as $skill) {
-                $currentLevel = $skill->Competency && $this->Student instanceof \Slate\People\Student ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
+                $currentLevel = $skill->Competency->getCurrentLevelForStudent($this->Student);
                 $demoSkillRating = $demoSkillIds[$skill->ID] ? $demoSkillIds[$skill->ID]->DemonstratedLevel : null;
 
                 $taskSkills[] = array_merge($skill->getData(), [
@@ -199,7 +199,7 @@ class StudentTask extends \VersionedRecord
 
         if ($this->Skills) {
             foreach ($this->Skills as $skill) {
-                $currentLevel = $skill->Competency && $this->Student instanceof \Slate\People\Student ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
+                $currentLevel = $skill->Competency->getCurrentLevelForStudent($this->Student);
                 $demoSkillRating = $demoSkillIds[$skill->ID] ? $demoSkillIds[$skill->ID]->DemonstratedLevel : null;
 
                 $taskSkills[] = array_merge($skill->getData(), [
--- a/php-classes/Slate/CBL/Tasks/StudentTask.php
+++ b/php-classes/Slate/CBL/Tasks/StudentTask.php
@@ -174,7 +174,7 @@ class StudentTask extends \VersionedRecord
 
         if ($this->Task && $this->Task->Skills) {
             foreach ($this->Task->Skills as $skill) {
-                $currentLevel = $skill->Competency ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
+                $currentLevel = $skill->Competency && $this->Student instanceof Slate\People\Student ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
                 $demoSkillRating = $demoSkillIds[$skill->ID] ? $demoSkillIds[$skill->ID]->DemonstratedLevel : null;
 
                 $taskSkills[] = array_merge($skill->getData(), [
@@ -188,7 +188,7 @@ class StudentTask extends \VersionedRecord
 
         if ($this->Skills) {
             foreach ($this->Skills as $skill) {
-                $currentLevel = $skill->Competency ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
+                $currentLevel = $skill->Competency && $this->Student instanceof Slate\People\Student ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
                 $demoSkillRating = $demoSkillIds[$skill->ID] ? $demoSkillIds[$skill->ID]->DemonstratedLevel : null;
 
                 $taskSkills[] = array_merge($skill->getData(), [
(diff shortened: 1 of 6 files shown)


php-classes/Slate/CBL/Tasks/StudentTask.php
```diff
@@ -188,7 +187,7 @@ public function getTaskSkills()
 
         if ($this->Skills) {
             foreach ($this->Skills as $skill) {
-                $currentLevel = $skill->Competency ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
+                $currentLevel = $skill->Competency && $this->Student instanceof \Slate\People\Student ? $skill->Competency->getCurrentLevelForStudent($this->Student) : null;
```
themightychris: @nbey this seems like a kind of hacky space to do this, how are StudentTask records getting creating with StudentID set to people that aren't students? This check should be done during validation of StudentTask. We shouldn't be guarding against corrupt data on the UI/reporting layer

themightychris (changes requested):

themightychris:
@nbey @ckunkel the conclusions we just came to:

- in the short term for this release
  - Remove the is-student checks
  - Delete non-student StudentTasks from production instances after a quick manual review and confirmation that they are junk data
  - Override validate for StudentTask to disallow `StudentID` getting pointed at a non-student person
- in the long term:
  - phase out using `Student` as a typehint on things that only need to be `Person` so that eventually any person in a school can participate in any CBL functionality

nbey:
@themightychris Can you review these changes please?

php-classes/Slate/CBL/Tasks/StudentTask.php
```diff
@@ -156,6 +156,18 @@ public function getValue($name)
         }
     }
 
+    protected static function validateStudent(\RecordValidator $RecordValidator, StudentTask $StudentTask, $options, $validator)
+    {
+        if (!$RecordValidator->hasErrors('Student')) {
+            $Student = $StudentTask->Student;
```
themightychris: @nbey avoid doing such variable aliasing in PHP

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -293,15 +293,36 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+                    me.doRateStudentTask(rec);
+                }
+            } else {
+                rec.reject();
+                if (response.failed.length) {
+                    validationErrors = response.failed[0].validationErrors;
+                    for (key in validationErrors) {
+                        if (validationErrors.hasOwnProperty(key)) {
+                            message.push(
+                                '<br>' + validationErrors[key] + '<br>'
```
themightychris: @nbey wrap each message in `<p>...</p>`

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -293,15 +293,36 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+                });
 
-            if (status === 're-assigned') {
-                me.doRateStudentTask(rec);
+                if (status === 're-assigned') {
+                    me.doRateStudentTask(rec);
+                }
+            } else {
+                rec.reject();
+                if (response.failed.length) {
```
themightychris: @nbey it is unsafe to assume that `response.failed` is defined, a connection failure could cause the request to fail with no response data. This should be `response.failed && response.failed.length`

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -394,13 +415,37 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+                me.getStudentTasksStore().load({
+                    id: studentTask.getId(),
+                    addRecords: true
+                });
+                Ext.toast(student.getFullName() + ' was assigned task: ' + task.get('Title'));
+            } else {
+                response = JSON.parse(request.getResponse().responseText);
+                message.push(
+                    'Failed to assign',
+                    '<b>'+rec.get('Task').Title+'</b>',
```
themightychris: @nbey use `<strong>` instead of `<b>` and wrap the message in a `<p>` instead of stacking `<br>`s to do the same thing

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -394,13 +415,37 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+
+            if (success) {
+                taskAssigner.close();
+                me.getStudentTasksStore().load({
+                    id: studentTask.getId(),
+                    addRecords: true
+                });
+                Ext.toast(student.getFullName() + ' was assigned task: ' + task.get('Title'));
+            } else {
+                response = JSON.parse(request.getResponse().responseText);
```
themightychris: @nbey use `Ext.decode` instead of `JSON.parse`, if the response had `application/json` as a content-type though the jarvus-apikit wrapper would have already decoded the response and populated `response.responseData`, you shouldn't need to parse json here

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -293,15 +293,36 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
         studentTask.set('TaskStatus', status);
 
-        me.doSaveStudentTask(studentTask, function(rec) {
-            taskRater.close();
-            me.getStudentTasksStore().load({
-                id: studentTask.getId(),
-                addRecords: true
-            });
+        me.doSaveStudentTask(studentTask, function(rec, success, request) {
+            var response = JSON.parse(request.getResponse().responseText),
```
themightychris: @nbey see the comment about `JSON.parse` in the below call to `doSaveStudentTask`

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -394,13 +415,37 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+            } else {
+                response = JSON.parse(request.getResponse().responseText);
+                message.push(
+                    'Failed to assign',
+                    '<b>'+rec.get('Task').Title+'</b>',
+                    'to',
+                    '<b>' + rec.get('Student').FirstName + ' ' + rec.get('Student').LastName + '</b>:',
+                    '<br><br>'
+                );
+                if (response.failed.length) {
```
themightychris: can't assume `response.failed` is defined here

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -414,13 +459,33 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
             studentTask.set('SkillIDs', me.getSkillsField().getSkills(false, true)); // returnRecords, idsOnly
-            return me.doSaveStudentTask(studentTask, function() {
-                me.getStudentTasksStore().load({
-                    id: studentTask.getId(),
-                    addRecords: true
-                });
-                taskEditor.close();
-                Ext.toast('Student task successfully updated.');
+            return me.doSaveStudentTask(studentTask, function(rec, success, request) {
+                var response = JSON.parse(request.getResponse().responseText),
```
themightychris: @nbey I've commented on the same code 3 times now, perhaps `me.doSaveStudentTask` should be handling reporting failures? Why else create a wrapper for that call within the controller if it's only going to turn two callbacks into one (which it doesn't need to do because `.save()` already has a `callback` option that gets called with a `success` boolean in either case)

themightychris (changes requested):

nbey:
@themightychris I moved the logic for handling errors into the save callback inside of `me.doSaveStudentTask`. 
I wasn't able to find the `responseData` shortcut that `jarvus-apikit` should append, nor the corresponding code in both `jarvus-apikit` & `emergence-apikit` repos, but I do recall using it before.. Any idea what could be going on with that?

themightychris:
@nbey here's the code path, see if you can get a breakpoint on it: https://github.com/JarvusInnovations/jarvus-apikit/blob/master/src/util/AbstractAPI.js#L103

nbey:
@themightychris I've been able to track down the issue, but I'm not sure how to properly fix.. It appears that the `autoDecode` config option is not being passed through, for various reasons. 
- When the request is created, inside of the `buildRequest` method in [Emergence.proxy.Records](https://github.com/JarvusInnovations/emergence-apikit/blob/master/src/proxy/Records.js) does not copy over the `autoDecode` config from the operation.
- Also, it appears that `Ext.data.Request` needs the `autoDecode` config to be added to it's `defaultConfig` object, which seems to be private/undocumented.

themightychris:
@nbey autoDecode doesn't need to be passed or copied from anywhere, the check is for `autoDecode !== false` so it is enabled by default when not set. autoDecode is on unless you explicitely pass `false`. There may be an issue with it not getting passed down somewhere (please file an issue in the relevant repo if you confirm this) but its omission should lead to it being active

themightychris (approved):

PR #369, opened 2017-02-01
Fix issues with rating tasks

nbey:
- [x] Add buttons for saving and resetting unsaved ratings.
- [x] Save ratings in batch
- [x] Allow removal of ratings
- [x] Verify competency promotion when rating skills
- [x] Automatically graduate students in competencies when editing an existing rating, that previously prevented a promotion, due to the average level of skill logs being insufficient.


themightychris:
V3 solves this a different way

PR #374, opened 2017-03-17
Release: slate-cbl v2.1.19

nbey:
- Update README installation instructions
- Fix skill evidence requirement import

### DIFF
--- a/site-root/cbl/import-maps.php
+++ b/site-root/cbl/import-maps.php
@@ -30,7 +30,7 @@ while ($mapsRow = $mapsCsv->getNextRow()) {
             \Debug::dumpVar($lastCompetency, false, 'creating competency');
         } elseif ($row['Type'] == 'Standard') {
             // handle numeric er values
-            if (is_numeric($row['ER'])) {
+            if (ctype_digit($row['ER'])) {
                 $demonstrationRequirements = [
                     'default' => $row['ER']
                 ];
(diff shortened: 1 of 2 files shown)


site-root/cbl/import-maps.php
```diff
@@ -29,12 +29,30 @@
...
+            if (is_numeric($row['ER'])) {
+                $demonstrationRequirements = [
+                    'default' => $row['ER']
+                ];
+            } else { // handle comma-delimited level:values pairs: i.e 9:2,10:3,default:4
+                $splitERs = explode(',', $row['ER']);
+                foreach ($splitERs as $er) {
+                    list($lvl, $total) = explode(':', $er);
+
+                    if (($lvl != 'default' && !is_numeric($lvl)) || !is_numeric($total)) {
```
themightychris: @nbey it would be better to use `ctype_digit` here. The benefit of `is_numeric` is that it will return true both for a number value and a string containing a number. In this case, since we're working with the output of explode, we can be certain that we're looking at strings. The benefit of `ctype_digit` is that it will only accept a string that contains digits, where's `is_numeric` will accept all sorts of weird stuff like negatives, decimals, exponential notation. In this case we *only* want integer strings so `ctype_digit` will give us a tighter test

I also think we should throw an exception here instead of just continuing. It would be better for the import to fail loudly if it finds something weird than to act like everything is cool and skip over data

themightychris (changes requested):

PR #376, opened 2017-03-30
GDrive Implementation

nbey:


php-classes/Slate/CBL/Tasks/Attachments/GoogleDocument.php
```diff
@@ -2,16 +2,33 @@
...
 namespace Slate\CBL\Tasks\Attachments;
 
-class GoogleDocument extends AbstractTaskAttachment
+class GoogleDocument extends Link
 {
     public static $fields = [
-        'ExternalID'
+        'FileID' => 'uint',
+        'RevisionID',
+        'Share' => [
```
themightychris: @nbey rename this to `ShareMethod`

php-classes/Slate/CBL/Tasks/Attachments/GoogleDocument.php
```diff
@@ -2,16 +2,33 @@
 
 namespace Slate\CBL\Tasks\Attachments;
 
-class GoogleDocument extends AbstractTaskAttachment
+class GoogleDocument extends Link
```
themightychris: @nbey why should this extend link? A google drive file doesn't have a URL field to store distinctly from its ID. It seems to me that it should continue extending `AbstractTaskAttachment`

themightychris: @nbey also, this class should be renamed to `GoogleDriveFile`

php-classes/Slate/CBL/Tasks/Attachments/GoogleDocument.php
```diff
@@ -2,16 +2,33 @@
...
+        'RevisionID' => [
             'validator' => 'string',
             'required' => true
         ]
     ];
+
+    public static $relationships = [
+        'File' => [
+            'type' => 'one-one',
+            'class' => \Google\File::class
```
themightychris: @nbey let's call this class `\Google\DriveFile`

php-classes/Google/API.php
```diff
@@ -0,0 +1,157 @@
...
+                'Content-Type: application/x-www-form-urlencoded'
+            ],
+            'post' => http_build_query([
+                'assertion' => JWT::encode($assertion, static::$privateKey, 'RS256'),
+                'grant_type' => 'urn:ietf:params:oauth:grant-type:jwt-bearer'
+            ])
+        ]);
+
+        if (empty($response['access_token'])) {
+            \Debug::dumpVar($response, false, 'api response');
```
themightychris: @nbey remove this

php-classes/Google/File.php
```diff
@@ -0,0 +1,84 @@
...
+    public static $fields = [
+        'ExternalIdentifier',
+        'OwnerEmail',
+        'Title',
+        'Type',
+        'Status' => [
+            'values' => ['normal', 'trashed', 'deleted'],
+            'default' => 'normal'
+        ],
+        'ParentDocumentIdentifier' => [
```
themightychris: @nbey rename to `ParentDriveID`

php-classes/Google/File.php
```diff
@@ -0,0 +1,84 @@
...
+{
+    public static $tableName = 'google_files';
+
+    public static $singularNoun = 'google file';
+    public static $pluralNoun = 'google files';
+
+    public $details;
+
+    public static $fields = [
+        'ExternalIdentifier',
```
themightychris: @nbey rename to `DriveID`

themightychris: @nbey add `'unique' => true` option

sencha-workspace/packages/slate-cbl/src/model/tasks/Attachment.js
```diff
@@ -13,46 +13,60 @@ Ext.define('Slate.cbl.model.tasks.Attachment', {
...
-            name: "ExternalID",
-            type: "string",
+            name: 'ExternalID',
+            type: 'string',
             allowNull: true
         },
         {
-            name: "URL",
-            type: "string",
+            name: 'RevisionID',
```
themightychris: @nbey exclude the RevisionID field from client-side models

sencha-workspace/packages/slate-cbl/src/view/AttachmentsList.js
```diff
@@ -82,31 +82,46 @@ Ext.define('Slate.cbl.view.AttachmentsList', {
...
                         },
-                        items: [
-                            {
-                                text: 'View Only',
-                                checked: true
+                        items: [{
+                            items: [{
+                                boxLabel: 'View Only',
+                                inputValue: 'view-only',
+                                checked: Boolean(record.get('Share') == 'view-only')
```
themightychris: @nbey did some linter make you use this explicit cast? `==` should always return boolean inherently

themightychris (changes requested):

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -458,8 +468,276 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+        }
+        // check for google token in cookies
+        gapi.load('auth:picker:client', {
+            callback: function() {
+                // TODO: check if token is any good
+                if (token) {
+                    return me.doOpenFilePicker(token);
+                }
+
+                me.doGoogleAuthentication();
```
themightychris: @nbey doOpenFilePicker should get passed as a callback to this method rather than being hard-coded into the chain of functions that follows authentication

We're going to want to break out all these functions to a sharable singleton class in the slate-cbl package so that other UIs (like the student submission UI we'll implement by the end of this project) can use the same code. Think about how to implement a clean, resuable set of methods there that could be called from here and chained together here through callback parameters

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -458,8 +468,276 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+        };
+
+        // confirm authentication is within google apps domain
+        gapi.client.request({
+            path: '/drive/v3/about',
+            params: {
+                fields: 'user'
+            }
+        }).
+        then(_verifyUserIsWithinDomain);
```
themightychris: @nbey inline the function instead of storing it in a variable to keep things easier to follow unless there are multiple paths to that function that necessitate making it reusable

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -458,8 +468,276 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+            gapi.client.request({
+                path: '/drive/v3/files/'+doc[google.picker.Document.ID]+'/permissions',
+                method: 'GET',
+                params: {
+                    'access_token': googleAppsToken
+                }
+            }).
+            then(_getGoogleFileOwner).
+            then(_getGoogleFileOwnerEmail).
+            then(_doVerifyGoogleFileOwner);
```
themightychris: @nbey again, inline functions if they're not being reused so we can see all the code in sequence

themightychris (changes requested):

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -458,8 +468,276 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+        _userIsNotWithinDomain = function(errorMessage) {
+            Ext.Msg.alert('Error', errorMessage);
+            return;
+        };
+
+        _verifyUserIsWithinDomain = function(response) {
+            console.info('loaded user', response.result);
+
+            if (!response.result || !response.result.user || !response.result.user.emailAddress.match(googleAppsEmailRegex)) {
+                return _userIsNotWithinDomain('Account must be under the correct domain. ' + googleAppsDomain);
```
themightychris: @nbey reformat this message so it reads `'Please sign in with an account under slatedemo.com'`

themightychris: @nbey after this dialog is shown, clicking "attachment" again jumps right back to it without letting the user sign in again. See if there is a client api call you can make to "deauth" the user from the current js api session, ideally without signing them out of google. If you reload the page you can pick an account again so there should be a call in the js API to achieve the same thing

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
```diff
@@ -458,8 +468,276 @@ Ext.define('SlateTasksTeacher.controller.Dashboard', {
...
+                // return false;
+            }
+
+            tokenExpiration = new Date(authResult.expires_at * 1000);
+
+            Ext.util.Cookies.set('googleAppsToken', authResult.access_token, tokenExpiration, '/');
+            Ext.util.Cookies.set('googleAppsPermissionId', response.result.user.permissionId, tokenExpiration, '/');
```
themightychris: @nbey I don't think we should store the permissionId in a cookie, only the token. If the user comes back with a token, we should hit whatever API provides us the permissionId first to both verify that the token is valid and re-obtain the permission id for the current page load

themightychris (changes requested):

sencha-workspace/SlateTasksTeacher/app/controller/Dashboard.js
themightychris: looks like reloading the page doesn't reset it either, we need to find a js api call we can make to reset auth, ideally without logging out the user's session

php-classes/Google/DriveFile.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+namespace Google;
+
+use \Slate\CBL\Tasks\Attachments\GoogleDriveFile;
+use \Slate\CBL\Tasks\Task;
+use \Slate\CBL\Tasks\StudentTask;
```
themightychris: @nbey code in the `Google` namespace should not be relying on code in the `Slate` namespace. Any `Slate`-dependent code should be in the `Slate` namespace. It doesn't look like any of these are actually used in this file anyway.

Also, `use` statements don't need to start with `\`, they are inherently absolute.

php-classes/Google/DriveFile.php
```diff
@@ -0,0 +1,108 @@
...
+                $details = $File->getGoogleFileDetails();
+                $File->Title = $File->details['name'];
+            }
+        } catch (\Exception $e) {
+            $valiadtor->addError('Title', 'Title is missing or invalid.');
+        }
+
+        if (!\Validators::string($File->Title)) {
+            $validator->addError('Title', 'Title is missing or invalid.');
+        }
```
themightychris: @nbey put `Validators`, `Exception`, and `RecordValidator` in `use` statements at the top and drop the leading `\` throughout the class body.

php-classes/Google/DriveFile.php
```diff
@@ -0,0 +1,108 @@
...
+        } catch (\Exception $e) {
+            $valiadtor->addError('Type', 'Type is missing or invalid.');
+        }
+
+        if (!\Validators::string($File->Type)) {
+            $validator->addError('Type', 'Type is missing or invalid.');
+        }
+    }
+
+    public static function validateTitle(\RecordValidator $validator, \Google\DriveFile $File)
```
themightychris: @nbey `DriveFile` can be referenced just as `DriveFile` since it's in the same namespace. Further I think you could just use `self` or `static` here

themightychris (changes requested):

php-classes/Slate/CBL/Tasks/TasksRequestHandler.php
```diff
@@ -377,7 +377,7 @@ public static function updateGoogleFilePermissions(Task $Record)
                                 'File' => $DriveFile,
                                 'Context' => $StudentTask,
                                 'ParentAttachment' => $attachment,
-                                'RevisionID' => 1
+                                'FileRevisionID' => 1
```
themightychris: @nbey do we not have access to the actual revision ID? Does `1` pass string validation?

php-classes/Google/Drive/UserChangeMonitorRequestHandler.php
```diff
@@ -0,0 +1,163 @@
...
+                if (substr($name, 0, 5) == 'HTTP_') {
+                    $headers[str_replace(' ', '-', ucwords(strtolower(str_replace('_', ' ', substr($name, 5)))))] = $value;
+                }
+            }
+            return $headers;
+        }
+        $requestHeaders = getallheaders();
+
+        $channelId = $requestHeaders['X-Goog-Channel-Id'];
+        $resourceState = $requestHeaders['X-Goog-Resource-State'];
```
themightychris: @nbey why all this reformatting of all the headers instead of just
```
$channelId = $_SERVER['HTTP_X_GOOGL_CHANNEL_ID'];
$resourceState = $_SERVER['HTTP_X_GOOG_RESOURCE_STATE'];
```

php-classes/Slate/CBL/Tasks/StudentTasksRequestHandler.php
```diff
@@ -162,75 +165,46 @@ public static function handleStudentTaskSubmissionRequest()
...
-            $originalAttachments = $Record->Attachments;
-            $originalAttachmentIds = array_map(function($s) {
-                return $s->ID;
-            }, $originalAttachments);
 
-            $failed = [];
-            $attachmentIds = [];
-            $attachments = [];
-            $defaultAttachmentClass = AbstractTaskAttachment::$defaultClass;
+        $defaultAttachmentClass = AbstractTaskAttachment::class;
```
themightychris: @nbey `AbstractTaskAttachment` isn't actually declared as an `abstract` class (although it should be or should be named something else). If it really is an abstract class, we can't have an instance of it, and so it shouldn't be a default. An attachment without class=Link or class=GoogleDriveFile is invalid and should generate a validation error or exception, and the abstract class should be actually declared as such so that any attempt to instantiate one generates a proper error instead of just silent weirdness

php-classes/Slate/CBL/Tasks/StudentTasksRequestHandler.php
```diff
@@ -162,75 +165,46 @@ public static function handleStudentTaskSubmissionRequest()
...
-
-            foreach ($newSkillIds as $newSkill) {
-                if (!$taskSkill = TaskSkill::getByWhere(['TaskID' => $Record->Task->ID, 'SkillID' => $newSkill])) { // check if skill is attached to related task first
-                    StudentTaskSkill::create([
-                        'StudentTaskID' => $Record->ID,
-                        'SkillID' => $newSkill
-                    ], true);
+                if ($Attachment instanceof Attachments\GoogleDriveFile) {
+                    if (!$Attachment->File) {
+                        if (!$File = \Google\DriveFile::getByWhere(['DriveID' => $attachmentData['File']['DriveID']])) {
```
themightychris: @nbey `getByField` would be more appropriate here and get to benefit from caching

themightychris (changes requested):

php-classes/Slate/CBL/Tasks/TasksRequestHandler.php
```diff
@@ -163,11 +164,74 @@ public static function handleTaskAssigneesRequest(ActiveRecord $Record)
...
+            $Record->Attachments = $attachments;
+        }
+        // update student tasks
+        if (isset($requestData['Assignees'])) {
+            $studentTasks = [];
+            foreach ($requestData['Assignees'] as $assigneeId) {
+                if (!$StudentTask = StudentTask::getByWhere(['StudentID' => $assigneeId, 'TaskID' => $Record->ID])) {
+                    $StudentTask = StudentTask::create([
+                        'StudentID' => $assigneeId,
+                        'SectionID' => $_GET['SectionID'],
```
themightychris: @nbey Why does this need to come in through the side as a GET param? If it's part of the record it should come in from `$requestData` with the record's other fields. If it's part of the saved record, it seems like the client code could just as easily declare it as a model field and set it before calling `record.save()` as easily as attaching custom request-level params to the save operation?

PR #378, opened 2017-06-25
Update Ext version to 6.2.0

nbey:
- Update submodule
- Update framework within workspace / apps
- Update Sencha CMD version to 6.2.0.103
- Fix minor issue within `SlateTasksTeacher` app


themightychris:
@nbey the package submodules also need to all be moved to an appropriate branch

themightychris (approved):

PR #383, opened 2017-08-12
Baselines and growth

themightychris:


php-classes/Slate/CBL/Competency.php
```diff
@@ -136,6 +136,16 @@ public function getSkillIds($forceRefresh = false)
         return $skillIds;
     }
 
+    private $totalSkills;
+    public function getTotalSkills($forceRefresh = false)
+    {
+        if (!empty($this->totalSkills)) {
```
themightychris: the difference between `if (!empty($this->totalSkills))` and `if ($this->totalSkills)` is that !empty additionally checks if the variable is defined, and it's impossible for this not to be defined since it's a declared member, so using `empty` here is superfluous

php-classes/Slate/CBL/Competency.php
```diff
@@ -136,6 +136,16 @@ public function getSkillIds($forceRefresh = false)
...
+            return $this->totalSkills;
+        }
+
+        return $this->totalSkills = count($this->getSkillIds());
+    }
```
themightychris: @nbey better pattern is to frame these functions like:

```php
    public function getTotalSkills($forceRefresh = false)
    {
        if ($this->totalSkills === null) {
            $this->totalSkills = count($this->getSkillIds());
        }

        return $this->totalSkills;
    }
```

(`=== null` is needed because `0` is a legit value that would pass `== null`)

php-classes/Slate/CBL/Competency.php
```diff
@@ -247,17 +263,18 @@ public function getCompletionForStudent(Student $Student, $level = null, $demons
...
                     Skill::$tableName
                 ];
 
                 $completion = DB::oneRecord($query, $parameters);
 
                 // cast strings to floats
                 return [
-                    'currentLevel' => $currentLevel,
+                    'currentLevel' => $StudentCompetency->Level,
+                    'baselineRating' => $StudentCompetency->BaselineRating ? floatval($StudentCompetency->BaselineRating) : null,
```
themightychris: @nbey activerecord should be in charge returning fields as their correct PHP type. If `decimal` doesn't come back as a float that should be added to activerecord the same way it handles ints

php-classes/Slate/CBL/Competency.php
```diff
@@ -271,8 +288,9 @@ public function getCompletionForStudent(Student $Student, $level = null, $demons
         return [
             'demonstrationsLogged' => 0,
             'demonstrationsComplete' => 0,
+            'baselineRating' => $StudentCompetency && $StudentCompetency->BaselineRating ? floatval($StudentCompetency->BaselineRating) : null,
             'demonstrationsAverage' => null,
-            'currentLevel' => $currentLevel
+            'currentLevel' => null
```
themightychris: @nbey current level still needs to be returned here if it's available

php-classes/Slate/CBL/Competency.php
```diff
@@ -271,8 +288,9 @@ public function getCompletionForStudent(Student $Student, $level = null, $demons
         return [
             'demonstrationsLogged' => 0,
             'demonstrationsComplete' => 0,
+            'baselineRating' => $StudentCompetency && $StudentCompetency->BaselineRating ? floatval($StudentCompetency->BaselineRating) : null,
```
themightychris: @nbey this can just be `$StudentCompetency ? $StudentCompetency->BaselineRating : null` once the ActiveRecord returns the correct type on its own

php-classes/Slate/CBL/Demonstrations/Demonstration.php
```diff
@@ -112,10 +112,17 @@ public function getCompetencyCompletions()
 
         $completions = [];
         foreach ($competencies AS $Competency) {
-            $completion = $Competency->getCompletionForStudent($this->Student);
-            $completion['StudentID'] = $this->StudentID;
-            $completion['CompetencyID'] = $Competency->ID;
-            $completions[] = $completion;
+            $StudentCompetency = StudentCompetency::getCurrentForStudent($this->Student);
+
+            $completions[] = [
```
themightychris: @nbey let's add a method `getCompletion` to `StudentCompetency` that handles returning this data structure since it's a standard of sorts and we're repeating it in a bunch of places.

php-classes/Slate/CBL/Demonstrations/StudentDashboardRequestHandler.php
```diff
@@ -122,10 +122,29 @@ public static function handleCompletionsRequest()
...
 
         foreach ($competencies AS $Competency) {
-            $completions[] = array_merge([
-                'StudentID' => $Student->ID,
-                'CompetencyID' => $Competency->ID
-            ], $Competency->getCompletionForStudent($Student));
+            $StudentCompetency = StudentCompetency::getCurrentForStudent($Competency, $Student);
+
+            if ($StudentCompetency) {
+                $completions[] = [
```
themightychris: @nbey use `getCompletion()`

php-classes/Slate/CBL/Demonstrations/Demonstration.php
themightychris: wire up a dynamic field to it too

event-handlers/Slate/CBL/Demonstrations/DemonstrationSkill/afterRecordSave/50_complete-competency.php
```diff
@@ -20,6 +20,11 @@
         'StudentID' => $Student->ID,
         'CompetencyID' => $Competency->ID,
         'Level' => $completion['currentLevel'] + 1,
-        'EnteredVia' => 'graduation'
+        'EnteredVia' => 'graduation',
+        'BaselineRating' => $completion['demonstrationsAverage']
     ], true);
+} elseif (empty($completion['baselineRating'])) {
+    if ($StudentCompetency = $Competency->getStudentCompetency($Student)) {
+        $StudentCompetency->calculateStartingRating();
```
themightychris: `calculateStartingRating` is supposed to return a value, not mutate state itself. The writing and saving if it comes back non-null should happen here. The verb "calculate" doesn't sound like a command to mutate state

event-handlers/Slate/CBL/Demonstrations/DemonstrationSkill/afterRecordSave/50_complete-competency.php
```diff
@@ -20,6 +20,11 @@
         'StudentID' => $Student->ID,
         'CompetencyID' => $Competency->ID,
         'Level' => $completion['currentLevel'] + 1,
-        'EnteredVia' => 'graduation'
+        'EnteredVia' => 'graduation',
+        'BaselineRating' => $completion['demonstrationsAverage']
     ], true);
+} elseif (empty($completion['baselineRating'])) {
+    if ($StudentCompetency = $Competency->getStudentCompetency($Student)) {
```
themightychris: `getStudentCompetency` doesn't appear to exist, also this can be consolidated into the if above:

```php
} elseif (empty($completion['baselineRating']) && ($StudentCompetency = StudentCompetency::getCurrentForStudent($Competency, $Student))) {
```

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+            '  FROM `%s` '.
+            ' WHERE CompetencyID = %u '.
+            'HAVING DemonstrationsRequirements > 0 ',
+            [
+                $this->Level,
+                Skill::$tableName,
+                $this->Competency->ID
+            ]
+        );
+        $validSkillIds = array_column($validSkills, 'ID');
```
themightychris: @nbey let's just foreach through `$this->Competency->Skills` and have that build up `$validSkillsId`, we're not gaining any efficiency doing all that json stuff in MySQL

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+                DemonstrationSkill::$tableName,
+                Demonstration::$tableName,
+                join(', ', $validSkillIds),
+                $this->Level,
+                $this->StudentID
+            ]
+        );
+
+        // check if each skill has been rated atleast once for the current level
+        if ((int)$ratedSkills !== count($validSkills)) {
```
themightychris: This check isn't sufficient, there could be ratings for skills with 0 demonstrations required. Use `DB::valuesTable` to get an array of totals indexed by skill ID, and then foreach through `$validSkillIds` and return null when you encounter one where `empty($skillRatings[$skillId])`

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+
+        // get the first rating (by date) for each skill
+        $demonstrationSkillRatings = DB::allValues(
+            'skillRatings',
+            'SELECT ds.DemonstratedLevel as skillRatings'.
+        	'  FROM `%1$s` ds '.
+    		'  JOIN `%2$s` d '.
+    		'    ON d.ID = ds.DemonstrationID '.
+    		' INNER JOIN ( '.
+                    'SELECT MIN(demoskill.Created) as Created, SkillID '.
```
themightychris: Created is when the rating was entered, we want the date of the demonstration.

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+            [
+                DemonstrationSkill::$tableName,
+                Demonstration::$tableName,
+                $this->Level,
+                join(', ', $validSkillIds),
+                $this->StudentID
+            ]
+        );
+
+        $this->BaselineRating = array_sum($demonstrationSkillRatings) / count($validSkills);
```
themightychris: return the calculation, don't mutate state

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+                    $this->StudentID,
+                    join(', ', $this->Competency->getSkillIds()),
+                    $this->Level
+                ]
+            );
+
+            return $this->demonstrationData = $data;
+        } catch (\Exception $e) {
+            throw $e;
+            return [];
```
themightychris: The return line will never run, what's the point of catching all exceptions and rethrowing them? Shouldn't this be catching TableNotFoundException and discarding it to return an empty array?

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+                Skill::$tableName,
+                $this->Competency->ID
+            ]
+        );
+        $validSkillIds = array_column($validSkills, 'ID');
+
+        $ratedSkills = DB::oneValue(
+
+            'SELECT SUM(IF(skillRatings.total > 0, 1, 0)) as totalRatings '.
+            '  FROM ( '.
```
themightychris: @nbey there's a mix of SQL indentation styles going on in this new function, you can see the preferred style here: https://github.com/EmergencePlatform/emergence-book/blob/master/recipes/make-a-model-searchable-by-tags.md

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+
+    public function calculateStartingRating($autoSave = true)
     {
-        $completion = $Competency->getCompletionForStudent($Student);
+        $validSkills = DB::allRecords(
+            'SELECT ID, '.
+            ' CAST( '.
+                'REPLACE( '.
+            '       IFNULL ( '.
+            '           JSON_EXTRACT(DemonstrationsRequired, CONCAT(\'$."\', %u, \'"\')), '.
```
themightychris: this CONCAT doesn't need to be done in SQL

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        'effectiveDemonstrationsData' => [
+            'getter' => 'getEffectiveDemonstrationsData'
+        ]
+
+    ];
+
+    public function calculateStartingRating($autoSave = true)
     {
-        $completion = $Competency->getCompletionForStudent($Student);
+        $validSkills = DB::allRecords(
```
themightychris: @nbey read but ignore most of my other feedback in this function, instead start it off with something like this:

```php
$ratings = DB::valuesTable(
    'SkillID',
    'Rating',
    '
    SELECT S.ID AS SkillID,
        IFNULL(
            JSON_EXTRACT(DemonstrationsRequired, \'$."%4$u"\'),
            JSON_EXTRACT(DemonstrationsRequired, "$.default")
        ) AS DemonstrationsRequirements,
        (
            SELECT DS.DemonstratedLevel
              FROM `%3$s` DS
              JOIN `%2$s` D
                ON D.ID = DS.DemonstrationID
             WHERE DS.SkillID = S.ID
               AND DS.TargetLevel = %4$u
               AND DS.DemonstratedLevel > 0
               AND D.StudentID = %5$u
             ORDER BY D.Demonstrated, D.ID
             LIMIT 1
        ) AS Rating
    FROM `%1$s` S
    WHERE S.CompetencyID = %6$u
    HAVING DemonstrationsRequirements > 0
    ',
    [
        Skill::$tableName, // %1$s
        Demonstration::$tableName, // %2$s
        DemonstrationSkill::$tableName, // %3$s
        'Level' => 9, // %4$u
        'StudentID' => 23, // %5$u
        'CompetencyID' => 1 // %6$u
    ]
)
```

This requires CBL-319 be done to eliminate the string->int conversion

themightychris: This will return null values where there is no first rating, so after this you just want to do:

```php
$ratingsCount = count(array_filter($ratings));
if ($ratingsCount != count($ratings) {
    return null;
}

return array_sum($ratings) / $ratingsCount;;
```

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+
+        if ($autoSave === true) {
+            $this->save();
+        }
+    }
+
+    private $demonstrationData;
+    public function getDemonstrationData()
+    {
+        if (!empty($this->demonstrationData)) {
```
themightychris: Use `!== null` because an empty array will keep failing `!empty` despite it representing a possible result that shouldn't keep getting refetched

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+                'SkillID',
+                'SELECT '.
+                '    DemonstrationSkill.*, '.
+                '    Demonstration.Demonstrated AS DemonstrationDate '.
+                '  FROM `%s` DemonstrationSkill '.
+                '  JOIN (SELECT ID, Demonstrated FROM `%s` WHERE StudentID = %u) Demonstration '.
+                '    ON Demonstration.ID = DemonstrationSkill.DemonstrationID '.
+                ' WHERE DemonstrationSkill.SkillID IN (%s) '.
+                '   AND DemonstrationSkill.TargetLevel = %u '.
+                ' ORDER BY SkillID, DemonstrationDate',
```
themightychris: order by DemonstrationID after DemonstartionDate so that multiple demonstrations on the same date get sorted consistently by order entered

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        } catch (\Exception $e) {
+            throw $e;
+            return [];
+        }
+    }
+
+    private $effectiveDemonstrationsData;
+    public function getEffectiveDemonstrationsData()
+    {
+        if (!empty($this->effectiveDemonstrationsData)) {
```
themightychris: `!== null`

php-classes/Slate/CBL/Competency.php
```diff
@@ -293,18 +311,4 @@ public static function getAllByListIdentifier($identifier)
...
-                    $Student->ID,
-                    $this->ID
-                ]
-            );
-        } catch (TableNotFoundException $e) {}
-
-        return $level ? intval($level) : null;
-    }
-}
+}
```
themightychris, line 214: `getCompletionForStudent` should also be getting deleted from this class

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+
+        $demonstrationsData = $this->getDemonstrationData();
+
+        foreach ($demonstrationsData as $skillId => &$demonstrationData) {
+            $sort = [];
+            foreach ($demonstrationData as $key => $row) {
+                $sort[$key]  = $row['DemonstratedLevel'];
+            }
+            // sort demonstrations by highest level
+            array_multisort($sort, SORT_DESC, $demonstrationData);
```
themightychris: I'm having trouble following what's going on here with `$sort` and `array_multisort`

You just need
```php
protected static function sortEffectiveDemonstrations($a, $b) {
    if ($a['DemonstratedLevel'] == $b['DemonstratedLevel']) {
        return 0;
    }
    return ($a['DemonstratedLevel'] < $b['DemonstratedLevel']) ? -1 : 1;
}
// ...
    foreach ($demonstrationsData as $skillId => &$demonstrationData) {
        uasort($demonstrationData, [__CLASS__,  'sortEffectiveDemonstrations']);
    }
```

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        }
+
+        return $this->effectiveDemonstrationsData = $demonstrationsData;
+    }
+
+    private $demonstrationsLogged;
+    public function getDemonstrationsLogged()
+    {
+
+        if (!empty($this->demonstrationsLogged)) {
```
themightychris: `0` is a valid value, use `!== null` and get rid of blank line

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        }
+
+        $effectiveDemonstrationsData = $this->getEffectiveDemonstrationsData();
+        $count = 0;
+        foreach ($effectiveDemonstrationsData as $skillId => $demonstrationData) {
+            $Skill = Skill::getByID($skillId);
+            $skillCount = 0;
+            foreach ($demonstrationData as $demonstration) {
+                if (empty($demonstration['Override'])) {
+                    if (!empty($demonstration['DemonstratedLevel'])) {
```
themightychris: combine these two ifs

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+                        $skillCount++;
+                    }
+                }
+            }
+            // ?
```
themightychris: ?

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+            $count += min($skillCount, $Skill->getDemonstrationsRequiredByLevel($this->Level));
```
themightychris: effective demonstrations is already capped by required

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        foreach ($effectiveDemonstrationsData as $skillId => $demonstrationsData) {
+            foreach ($demonstrationsData as $demonstration) {
+                if (empty($demonstration['Override'])) {
+                    $totalScore += $demonstration['DemonstratedLevel'];
+                }
+            }
+        }
+
+        if ($this->getDemonstrationsLogged()) {
+            $average = $totalScore / $this->getDemonstrationsLogged();
```
themightychris: just set `$this->demonstrationsAverage`

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+
+    private $demonstrationsComplete;
+    public function getDemonstrationsComplete()
+    {
+        if (!empty($this->demonstrationsComplete)) {
+            return $this->demonstrationsComplete;
+        }
+
+        $effectiveDemonstrationsData = $this->getEffectiveDemonstrationsData();
+        $count = 0;
```
themightychris: just use `$this->demonstrationsComplete`, initialize it here and return it, we don't need to worry about simultaneous access or anything like that

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        'demonstrationsAverage' => [
+            'getter' => 'getDemonstrationsAverage'
+        ],
+        'demonstrationData' => [
+            'getter' => 'getDemonstrationData'
+        ],
+        'effectiveDemonstrationsData' => [
+            'getter' => 'getEffectiveDemonstrationsData'
+        ]
+
```
themightychris: blank line

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        if (empty($this->BaselineRating)) {
+            return null;
+        }
+
+        $demonstrationData = $this->getDemonstrationData();
+
+        $totalSkills = $this->Competency->getTotalSkills();
+
+        $growthData = array_map(function($demonstrations, $skillId) {
+            if (count($demonstrations) === 1 && empty($this->BaselineRating)) {
```
themightychris: no need to use `empty` on a model field, it's not at risk of being undefined

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+                return null;
+            }
+
+            $lastRating = end($demonstrations);
+            if ($this->BaselineRating) {
+                return $lastRating['DemonstratedLevel'] - floatval($this->BaselineRating);
```
themightychris: `floatval` won't be needed here after the model field is fixed to return correct type

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+            } else {
+                $firstRating = reset($demonstrations);
+
+                return $lastRating['DemonstratedLevel'] - $firstRating['DemonstratedLevel'];
+            }
+
+        }, $demonstrationData, array_keys($demonstrationData));
```
themightychris: `$skillId` isn't being used in this map function, so we don't need to pass keys in (which causes the returned array to lose its key=>value association)

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+
+        array_filter($growthData);
```
themightychris: @nbey array_filter returns a new array it does not modify the passed array. Wrap it around the array_map result above

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+        $totalGrowthSkills = count($growthData);
+
+        if (count($growthData) * 2 < $totalSkills) {
+            return null;
+        }
+
+        return array_sum($growthData) / $totalSkills;
```
themightychris: needs to be memoized

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +63,334 @@ class StudentCompetency extends \ActiveRecord
...
+    }
+
+
+    public static function getCurrentForStudent(Competency $Competency, Student $Student)
+    {
+        return static::getByWhere(['StudentID' => $Student->ID, 'CompetencyID' => $Competency->ID], ['order' => ['Level' => 'DESC']]);
+    }
+
+    public static function isCurrentLevelComplete($Student, $Competency)
```
themightychris: needs type hints

php-classes/Slate/CBL/Competency.php
nbey: @themightychris Aren't we deprecating this method?

php-classes/Slate/CBL/StudentCompetency.php
nbey: @themightychris `DemonstrationDate` is actually a timestamp column, so there shouldn't be an issue with same day demonstrations.

php-classes/Slate/CBL/Competency.php
themightychris: Oh, yeah. As long as the new method doesn't have this regression... This whole thing can just get deleted

php-classes/Slate/CBL/StudentCompetency.php
themightychris: @nbey the column is a timestamp, but the current UI is only a date selector so the time components are mostly the same and we should secondarily sort by ID to ensure consistent sequencing. In the future we might see the time component getting specific values more frequently, especially when we start filling in demonstrations automatically from external systems or primary work completion

php-classes/Slate/CBL/Demonstrations/StudentDashboardRequestHandler.php
```diff
@@ -122,10 +122,21 @@ public static function handleCompletionsRequest()
...
-            $completions[] = array_merge([
-                'StudentID' => $Student->ID,
-                'CompetencyID' => $Competency->ID
-            ], $Competency->getCompletionForStudent($Student));
+            $StudentCompetency = StudentCompetency::getCurrentForStudent($Competency, $Student);
+
+            if ($StudentCompetency) {
+                $completions[] = $StudentCompetency->getCompletion();
+            } else {
+                $completions[] = [
```
themightychris: @nbey let's create a method `StudentCompetency::getBlankCompletion($Student, $Competency)` for returning this data structure and use it in the teacher request handler too so that as we add to / evolve this data structure there are only two places it is defined and they're both in the same class

php-classes/Slate/CBL/Demonstrations/StudentDashboardRequestHandler.php
```diff
@@ -122,10 +122,21 @@ public static function handleCompletionsRequest()
         $completions = [];
 
         foreach ($competencies AS $Competency) {
-            $completions[] = array_merge([
-                'StudentID' => $Student->ID,
-                'CompetencyID' => $Competency->ID
-            ], $Competency->getCompletionForStudent($Student));
+            $StudentCompetency = StudentCompetency::getCurrentForStudent($Competency, $Student);
```
themightychris: @nbey let's flip the parameter order for getCurrentForStudent so that it matches the name of the class and the order of the index

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +64,349 @@ class StudentCompetency extends \ActiveRecord
...
+            'baselineRating' => $this->BaselineRating,
+            'demonstrationsLogged' => $this->getDemonstrationsLogged(),
+            'demonstrationsComplete' => $this->getDemonstrationsComplete(),
+            'demonstrationsAverage' => $this->getDemonstrationsAverage()
+        ];
+    }
+
+    public function calculateStartingRating()
+    {
+        $validSkills = DB::allRecords(
```
themightychris: @nbey let's use `DB::table('ID', '...', [...])` and then `$validSkillIds = array_keys($validSkills)`

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +64,349 @@ class StudentCompetency extends \ActiveRecord
...
+    		   AND ds.Override = false
+            ',
+            [
+                DemonstrationSkill::$tableName,
+                Demonstration::$tableName,
+                $this->Level,
+                join(', ', $validSkillIds),
+                $this->StudentID
+            ]
+        );
```
themightychris: @nbey was there any issue with the code I developed here: https://github.com/SlateFoundation/slate-cbl/pull/383#discussion_r132990589

that got us everything we needed in one query

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +64,349 @@ class StudentCompetency extends \ActiveRecord
...
+                    Demonstration::$tableName,
+                    $this->StudentID,
+                    join(', ', $this->Competency->getSkillIds()),
+                    $this->Level
+                ]
+            );
+
+            return $this->demonstrationData = $data;
+        } catch (TableNotFoundException $e) {
+            return [];
```
themightychris: @nbey this code will result in the empty result of a table being missing not being cached, defeating the purpose of our strict null check to allow for empty arrays as legit values.

Use this pattern instead for all of these functions:

```php
if ($this->demonstrationData === null) {
    try {
        $this->demonstrationData = DB::arrayTable(
            // ...
        );
    } catch (TableNotFoundException $e) {
        $this->demonstrationData = [];
    }
}

return $this->demonstrationData;
```

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +64,349 @@ class StudentCompetency extends \ActiveRecord
...
+            foreach ($demonstrationData as $demonstration) {
+                if (empty($demonstration['Override']) && !empty($demonstration['DemonstratedLevel'])) {
+                    $skillCount++;
+                }
+            }
+            $count += $skillCount;
+        }
+
+        return $this->demonstrationsLogged = $count;
+
```
themightychris: blank line

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -53,16 +64,349 @@ class StudentCompetency extends \ActiveRecord
...
+            $skillCount = 0;
+            foreach ($demonstrationData as $demonstration) {
+                if (empty($demonstration['Override']) && !empty($demonstration['DemonstratedLevel'])) {
+                    $skillCount++;
+                }
+            }
+            $count += $skillCount;
+        }
+
+        return $this->demonstrationsLogged = $count;
```
themightychris: I mentioned this in the last round, there's no need to use this intermediate variables. Let's work on the variable we're trying to set directly as much as possible to reduce the cognitive load of reading the code

```php

if ($this->demonstrationsLogged === null) {
    $this->demonstrationsLogged = 0;
    foreach (...) {
        $this->demonstrationsLogged += $whatever;
    }
}

return $this->demonstrationsLogged;
```

php-classes/Slate/CBL/StudentCompetency.php
nbey: @themightychris No, for some reason I didn't see that.

PR #386, opened 2017-09-28
Baxter e nto ex

hallielarsson:


themightychris:
Deprecated by the rebased branch #392

PR #390, opened 2017-10-10
Tyler/teton fixes

jtwiest:
- Hide activity toggle from student task dashboard
- Hide task history from student task dashboard
- Add isLevelCompleted callable config
- Add minimumRatingOffset config to conditionally force a minimum demonstration level

### DIFF
--- a/php-classes/Slate/CBL/StudentCompetency.php
+++ b/php-classes/Slate/CBL/StudentCompetency.php
@@ -14,6 +14,7 @@ class StudentCompetency extends \ActiveRecord
 {
     public static $autoGraduate = true;
     public static $isLevelComplete;
+    public static $minimumRatingOffset = null;
 
     // ActiveRecord configuration
     public static $tableName = 'cbl_student_competencies';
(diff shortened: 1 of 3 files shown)


php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -13,6 +13,8 @@
 class StudentCompetency extends \ActiveRecord
 {
     public static $autoGraduate = true;
+    public static $isLevelComplete;
+    public static $minimumRatingOffset = null;
```
jtwiest: @themightychris There is a strict comparison against minimumRatingOffset on line 342, do I need to declare it as null here or can I simply leave the value blank?

themightychris: @jtwiest `null` is the default value for an uninitialized class property, explicitly setting it `= null` is not necessary

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -326,13 +328,35 @@ public function isLevelComplete()
...
+        }
+
+        // Require minimum average as offset from level
+        if ($minimumOffset !== null && $average < $this->Level + $minimumOffset) {
+            return false;
+        }
+
+        // Require all demonstrations have ratings above minimum
+        if (static::$minimumRatingOffset !== null) {
+            $minimumRating = $this->Level - static::$minimumRatingOffset;
```
themightychris, line 343: @jtwiest i renamed this variable to `$minimumRating`, it's best to avoid abbrevations

themightychris (approved):

PR #392, opened 2017-10-10
DO NOT MERGE: Baxter EN to EX customizations

themightychris:


themightychris:
baxter plugin for v3 is complete: https://github.com/baxter-cs/slate-cbl-baxter

PR #396, opened 2017-10-12
Filter task dashboard by student

jtwiest:
- Allow teachers to toggle student task sections
- Allow teachers to toggle active student in task dashboard


sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,48 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
         });
     },
 
+    onStudentSelectorSelect: function(combo, rec) {
+        var me = this,
+            studentCombo = me.getStudentSelectorCombo(),
+            params = me.getSearchParams(),
+            username;
+
+        if (!studentCombo) {
```
jtwiest: @themightychris This check was done as a precautionary measure. I don't see a reason why the studentCombo wouldn't be available but if it wasn't this would at least handle the error gracefully.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,48 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
+    onStudentSelectorBoxReady: function(combo) {
+        var params = this.getSearchParams();
+        combo.setValue(params.student);
+
+        // reset section state
+        this.redirectTo('section/all');
+    },
+
+    // parse active location.search params into object
+    getSearchParams: function() {
```
jtwiest: @themightychris I'm using this instead of just updating location.search.student = username (like we had it) to preserve the apiHost query param while developing. Furthermore, if there are any other query params that come into play at some point, this will help handle them gracefully.

sencha-workspace/SlateTasksStudent/app/view/AppHeader.js
```diff
@@ -19,6 +19,33 @@ Ext.define('SlateTasksStudent.view.AppHeader', {
...
         items: [{
+            xtype: 'combo',
+            itemId: 'student-selector',
+            name: 'StudentSelector',
+            fieldLabel: 'Student',
+            displayField: 'Username',
+            valueField: 'Username',
+            allowBlank: true,
+            queryParam: 'q',
+            margin: '0 10 0 0',
```
jtwiest: @themightychris this needed some space on the right side of it. I did some searching but I didn't see a standard class applied to this type of element to add spacing so I just added 10 pixels of margin to the right manually.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
themightychris: @jtwiest In this app there's no normal condition under which it should not exist. For things like this I'd prefer an abnormal condition to generate an error that our top-level error monitoring tools can catch and report than to silently dismiss it. If that component doesn't exist we're off the road and ignoring it isn't handling it. Also, putting guards in for things that shouldn't happen is a bottomless pit. IMHO it's better to just have a good system in place for reporting on unhandled errors. I would remove this guard

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
themightychris: @jtwiest good catch. I hate to break it to you, but `Ext.Object.fromQueryString`. Please switch to that, it handles more cases

sencha-workspace/SlateTasksStudent/app/view/AppHeader.js
```diff
@@ -19,6 +19,42 @@ Ext.define('SlateTasksStudent.view.AppHeader', {
...
         items: [{
+            xtype: 'combo',
+            itemId: 'student-selector',
+            name: 'StudentSelector',
+            fieldLabel: 'Student',
+            valueField: 'Username',
+            allowBlank: true,
+            queryParam: 'q',
+            margin: '0 10 0 0',
+            displayTpl: Ext.create('Ext.XTemplate',
```
themightychris: @jtwiest no classes should ever be instantiated during class-definition-time. This expression, `Ext.create(...)`, will get executed while this source file is being parsed, bypassing the framework's loader's ability to dynamically resolve dependencies and sequence loading, creating a race condition. 

The correct why to do this would be to define `displayTpl` as an array:
```
displayTpl: [
    '<tpl for=".">'
       '...'
    '</tpl>'
]
```

Components that have `*Tpl` configs use the [lookupTpl](http://docs.sencha.com/extjs/6.2.0/modern/Ext.Widget.html#method-lookupTpl) method to read them which handles instantiating a string/array into a `Ext.XTemplate` instance as needed.

But also, a better solution here is to add `Slate.model.person.Person` to the `requires` for this class at the top, and then set `model: 'Slate.model.person.Person'` for the store instead of `fields`, and then just use `displayField: 'FullName'` which utilized the `FullName` virtual field provided by the official [Person model](https://github.com/SlateFoundation/slate-core-data/blob/master/src/model/person/Person.js)

sencha-workspace/SlateTasksStudent/app/view/AppHeader.js
```diff
@@ -19,6 +19,42 @@ Ext.define('SlateTasksStudent.view.AppHeader', {
...
+                '<tpl for=".">',
+                    '{FirstName} {LastName}',
+                '</tpl>'
+            ),
+            tpl: Ext.create('Ext.XTemplate',
```
themightychris: @jtwiest see above, this isn't needed if you can just produce the desired format in the model

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,49 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
+        }
+
+        // reroute page to reflect selected student
+        params.student = username;
+        location.search = Ext.Object.toQueryString(params);
+    },
+
+    onStudentSelectorBoxReady: function(combo) {
+        var params = this.getSearchParams();
+        combo.getStore().load();
```
themightychris: @jtwiest `load()` is an asynchronous call, if you're relying on it this is a race condition. You could pass `{ callback: function() { ... } }` to it if you needed something to be done after it was loaded. You don't usually need to manually trigger combo box stores to load though, and we don't want it loading all people in the system with no query. Just set the value and let combobox worry about when it loads

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,49 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
         });
     },
 
+    onStudentSelectorSelect: function(combo, rec) {
```
themightychris: @jtwiest name these parameters `studentCombo, student`

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,49 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
+        var me = this,
+            studentCombo = me.getStudentSelectorCombo(),
```
themightychris: @jtwiest you already have `studentCombo`!

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,49 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
+            params = me.getSearchParams(),
+            username;
```
themightychris: @jtwiest you can initialize this to `student.get('Username')`, the event handler was passed the selected record

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,49 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
+        location.search = Ext.Object.toQueryString(params);
+    },
+
+    onStudentSelectorBoxReady: function(combo) {
+        var params = this.getSearchParams();
+        combo.getStore().load();        
+        combo.setValue(params.student);
+
+        // reset section state
+        this.redirectTo('section/all');
```
themightychris: shouldn't this be done when a student is selected instead? Redirecting during initial rendering is a bad idea generally

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -87,13 +90,22 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
         }
     },
 
 
     // controller templates method overrides
     onLaunch: function () {
         this.getDashboard().render('slateapp-viewport');
+
+        // hide student selector from students
+        if (SiteEnvironment.user && SiteEnvironment.user.AccountLevel == "User") {
```
themightychris: this needs to be prefixed with `window.SiteEnvironment &&` to avoid crashing if the global isn't set. `window.` doesn't need to be used after that if it is defined. And don't use double quotes for strings when single quotes are used everywhere else.

themightychris: Also, it's better to configure the component as `hidden: true` initially and then `setHidden(false)` if it should be shown, that way there is never a flash of it appearing.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -136,6 +148,49 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
+        if (username === params.student) {
+            return;
+        }
+
+        // reroute page to reflect selected student
+        params.student = username;
+        location.search = Ext.Object.toQueryString(params);
+    },
+
+    onStudentSelectorBoxReady: function(combo) {
```
themightychris: don't use `boxready` for this, what you're doing isn't actually dependent on the combobox having it's width/height calculated, this is a false causal chain and thar lie the path to chaos and jank. Do it up in `onLaunch` instead where you're already grabbing a reference to it. The component tree and API will be available as soon as you `this.getDashboard()` instantiates it on the first line of `onLaunch`

themightychris (changes requested):

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
jtwiest: @themightychris 👍  What top-level error monitoring tools are in place?

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
jtwiest: Awesome, I was looking for something like this but when I couldn't find it I nabbed a generic solution off stack overflow.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
jtwiest: @themightychris the window part makes sense.

I set the hidden logic how I did because if you're developing locally, window.SiteEnvironment.user doesn't exist meaning the dropdown will never actually be visible. Obviously this won't be a problem in production but it means you have to manually set hidden = false just to test this feature. Having a step like that in place to me seem like it could be prone to error or at least a dev accidentally committing the hidden flag accidentally.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
jtwiest: @themightychris I was following the lead set by `onSectionSelectorSelect` do you want me to update those param names as well?

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
jtwiest: @themightychris This is done to ensure if you switch from one student to another the previous section selection is removed. Ie if you have student 1 English 1 to student 2 you need to make sure it goes back to section/all in case student 2 is not a member of English 1.

I tried implementing this in onStudentSelectorSelect but if I try to reset the route and the change the query string I was getting an error.

sencha-workspace/SlateTasksStudent/app/view/AppHeader.js
jtwiest: @themightychris Interesting, I pulled that right off the docs but I suppose that was in a different context ie the combo was bing created vie Ext.create no the method we're using.

http://docs.sencha.com/extjs/6.0.2/classic/Ext.form.field.ComboBox.html
(Customized combobox example)

What you're suggesting makes sense though, thanks for the explanation.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
jtwiest: @themightychris I tried doing it without the load and it's not working b\c we're using a custom display tpl. In other words we're setting the value as the Username field but rendering the First / Last name fields.

I used the load callback method to populate this value instead.

jtwiest:
@themightychris Ok I made and the suggested changes with one exception regarding the implementation of `this.redirectTo('section/all');` on line 126.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
jtwiest: @themightychris I moved the section resetter to after the user choose a new student and the following error appears: 'TypeError: cannot read property 'mousedown' of null' 

You were correct in saying if you turn off the Pause on Exception flag (or don't have the inspector working at all) it'll work just fine however I still haven't found a way to remove this error from popping up in the first place.

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
themightychris: @jtwiest FYI when testing locally you can manipulate `window.SiteEnvironment` in the local `index.html`, just don't commit those changes

sencha-workspace/SlateTasksStudent/app/controller/Dashboard.js
```diff
@@ -87,13 +90,31 @@ Ext.define('SlateTasksStudent.controller.Dashboard', {
...
+        if (window.SiteEnvironment && SiteEnvironment.user && SiteEnvironment.user.AccountLevel != 'User') {
+            me.getStudentSelectorCombo().setHidden(false);
+        }
+
+        // load student combo and set value
+        studentCombo = me.getStudentSelectorCombo();        
+        studentCombo.getStore().load({ callback: function() {
+            var queryParams = Ext.Object.fromQueryString(location.search);
+            this.getStudentSelectorCombo().setValue(queryParams.student);
+        }, scope: me});
```
themightychris: @jtwiest don't use scope binding in sub-scopes, that's why we define `me` at the top, you can just use that throughout subscopes to reference the instance of the class defined by the current source file

themightychris: @jtwiest also `studentCombo` was already available in this scope

themightychris: @jtwiest also we shouldn't be loading the student combo if it's not visible to the current user

themightychris (approved):

PR #399, opened 2017-10-19
Improve styling for rating and level sliders

ryon:
The one line I'm not sure about is https://github.com/SlateFoundation/slate-cbl/compare/styling?expand=1#diff-2eb74777bbceddc93e313dfa920a135cR169 — the `parseInt()` seems a little fragile, but I'm not sure what a better way might be to check for the rating in the array.

**Gray selected colors and more appropriate font sizes:**
screenshot 2017-10-18 20 40 22
screenshot 2017-10-18 20 40 34

**&times; alignment:**
screenshot 2017-10-18 20 49 42


themightychris:
Merged into #400

PR #401, opened 2017-10-19
WIP: Component-based raters

themightychris:
This is a work-in-progress to break down the rater UI into sub components

- [ ] Make menu more like sliders':
screen shot 2017-10-18 at 9 57 24 pm


themightychris:
v3 supersedes this

PR #403, opened 2017-10-20
DO NOT MERGE: customized 1-4 scale ratings

themightychris:


themightychris:
v3 plugin infra resolves this without forks

PR #405, opened 2017-10-23
CBL-389

nbey:
- Update incomplete student tasks, when updating task.
- Reload student tasks after task saves
- Re-render student tasks in aggregrid, after save


themightychris:
v3 resolves this

PR #477, opened 2019-05-16
Restore Tasks Manager application

nbey:
- Restore Tasks Manager Application


themightychris:
@nbey could you add a basic "smoke test" like we have for the other apps as part of this PR (and rebase it on top of current `develop` with testing infra) ?

themightychris (changes requested):
@nbey lets comment out the multisearch bar and get this app building without it, then get this merged and take another pass on an improved filtering UI later

PR #500, opened 2020-01-15
Feat/growth interface

nbey:


themightychris:
So let's define a `Slate\CBL\IGrowthCalculator` interface with one method `public static function calculateGrowth(StudentCompetency $StudentCompetency)` and then add a static config `Slate\CBL\StudentCompetency::$growthCalculatorClass = MostRecentMinusFirst::class`

Define two classes that implement `IGrowthCalculator`:

- `Slate\CBL\MostRecentMinusFirst` (the original method)
- `Slate\CBL\PerformanceLevelMinusBaseline` (the new method)

Then the `StudentCompetency->getGrowth()` method will delegate the calculation to the configured class

For testing, let's move that data structure you put together out to a .json file under `fixtures` that we can load. We'll then need some way for the cypress test to know which calculation method is in effect...I'm not sure. Returning it in response data seems flakey, maybe the slate-cbl repo should come with a data fixture that matches against the default `MostRecentMinusFirst` calculator, and then the b21-skeleton repo that changes the default to `PerformanceLevelMinusBaseline` could override that file with the data here

This model will also lend itself well to phpunit testing, so we could also layer in phpunit tests that exercise both models within the slate-cbl repo.

For sake of moving this forward for now, let's go with moving the verification data structure to an external file, update its values to match the default behavior, and put an override in b21-skeleton for the new values

nbey:
@themightychris This is ready for another review.

php-classes/Slate/CBL/Calculators/Growth/MostRecentMinusFirst.php
```diff
@@ -0,0 +1,40 @@
...
+namespace Slate\CBL\Calculators\Growth;
+
+use Slate\CBL\StudentCompetency;
+
+class MostRecentMinusFirst implements IGrowthCalculator
+{
+    public static function calculateGrowth(StudentCompetency $StudentCompetency)
+    {
+        $demonstrationData = $StudentCompetency->getDemonstrationData();
+        $growthData = array_filter(array_map(function($demonstrations) use ($StudentCompetency) {
```
themightychris, line 12: @nbey you were missing the `use ($StudentCompetency)` bit here and the growth calcs were silently all very off. I'm going to add some vscode extension recommendations in `develop`, next time you sit down make sure you have them all installed and working, intelephense highlighted the issue for me otherwise I would have missed it too

themightychris (approved):
@nbey quite perfect work overall! in addition to the growth calculation issue I mentioned, some other minor issues I fixed included the database submodule not being set to your latest commit in the fixtures repo, and that you overwrote all the `Demonstrated` timestamps in the fixture data with the stock created value. We want the demonstrated times to be sequential for best testing I think, I changed them all to an ID-based sequence and logged the query I used for that in the commit message details

PR #503, opened 2020-01-21
fix: prevent parent task paradox

nbey:
- [x] prevent parent task from being set to itself
- [x] create migration script

### DIFF
--- a/php-classes/Slate/CBL/Tasks/Task.php
+++ b/php-classes/Slate/CBL/Tasks/Task.php
@@ -201,9 +201,12 @@ class Task extends \VersionedRecord
 
     protected static function validateParentTask($validator, Task $Task, $options)
     {
-        if (!$validator->hasErrors('ParentTask') && $Task->ParentTaskID) {
-            if ($Task->ParentTaskID === $Task->ID) {
-                $validator->addError('ParentTask', 'ParentTaskID and TaskID must be different.');
+        if (!$validator->hasErrors('ParentTaskID') && $Task->ParentTaskID) {
+            if ($Task->ParentTaskID == $Task->ID) {
+                $validator->addError('ParentTaskID', 'A task cannot be its own parent');
+
+                // clear immediately to prevent validation loop
+                $Task->ParentTaskID = null;
             }
         }
     }
--- a/php-classes/Slate/CBL/Tasks/Task.php
+++ b/php-classes/Slate/CBL/Tasks/Task.php
@@ -72,7 +72,10 @@ class Task extends \VersionedRecord
 
     public static $validators = [
         'Section' => 'require-relationship',
-        'Title'
+        'Title',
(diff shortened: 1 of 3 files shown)


php-classes/Slate/CBL/Tasks/Task.php
```diff
@@ -195,4 +198,16 @@ public static function delete($id)
...
     }
+
+    protected static function validateParentTask($validator, Task $Task, $options)
+    {
+        if (!$validator->hasErrors('ParentTaskID') && $Task->ParentTaskID) {
+            if ($Task->ParentTaskID == $Task->ID) {
+                $validator->addError('ParentTaskID', 'A task cannot be its own parent');
+
+                // clear immediately to prevent validation loop
+                $Task->ParentTaskID = null;
```
themightychris, line 209: @nbey FYI, I found this necessary when I tested the server-side rejection -- the response would hang indefinitely as validation went down an infinite recursion.

I tried for a while to update the JSON encoder or record validator to detect/avoid recursion, but no luck. This feels a bit hacky but stops validation from looping

sencha-workspace/SlateTasksTeacher/app/controller/Tasks.js
```diff
@@ -363,6 +363,12 @@ Ext.define('SlateTasksTeacher.controller.Tasks', {
                 Section: section.getData()
             }, task || null));
 
+            formPanel.getParentTaskField().getStore().filter({
+                id: 'exclude-self',
```
themightychris, line 367: @nbey FYI, I found that the initial implementation of this would keep adding an additional filter each time you opened the window. The [FilterCollection](https://docs.sencha.com/extjs/6.2.1/classic/Ext.util.FilterCollection.html) instance supports keying members by `id`, so this implementation will ensure that the same filter is updated each time. It also serves to ensure that the filter is removed when you open the "create" version of the form after opening an edit version, as the value checked for won't be a real ID that matches anything in the Tasks store

sencha-workspace/SlateTasksTeacher/app/controller/Tasks.js
```diff
@@ -363,6 +363,12 @@ Ext.define('SlateTasksTeacher.controller.Tasks', {
...
+                property: 'ID',
+                operator: '!=',
+                value: task.getId()
+            });
             formPanel.setTask(task);
```
themightychris, line 372: @nbey FYI, I also found it necessary that the filtering be updated BEFORE `setTask` loads the record into the form. Previously, if a parent task had been opened for editing immediately before the child task was, the parent task would be filtered out from the combo's values when the form loaded the child record and fail to match the existing ParentTaskID to the task record. This way, when the record is loaded, the available related values is already correct

themightychris (approved):

PR #505, opened 2020-01-23
fix: Performance Level Minus Baseline growth calculation

nbey:
- [x] fix: return null growth if there is no PL
- [x] fix: require baseline to be set
- [x] fix: use rounded baseline and pl

### DIFF
--- a/php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
+++ b/php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
@@ -8,8 +8,8 @@ class PerformanceLevelMinusBaseline implements IGrowthCalculator
 {
     public static function calculateGrowth(StudentCompetency $StudentCompetency)
     {
-        $baseline = $StudentCompetency->BaselineRating;
-        $performanceLevel = $StudentCompetency->getDemonstrationsAverage();
+        $baseline = round($StudentCompetency->BaselineRating, 1);
+        $performanceLevel = round($StudentCompetency->getDemonstrationsAverage(), 1);
 
         // performance level must > 0
         if (!$performanceLevel) {
--- a/php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
+++ b/php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
@@ -8,8 +8,8 @@ class PerformanceLevelMinusBaseline implements IGrowthCalculator
 {
     public static function calculateGrowth(StudentCompetency $StudentCompetency)
     {
-        $baseline = round($StudentCompetency->BaselineRating, 1);
-        $performanceLevel = round($StudentCompetency->getDemonstrationsAverage(), 1);
+        $baseline = $StudentCompetency->getBaselineRating();
+        $performanceLevel = $StudentCompetency->getDemonstrationsAverage();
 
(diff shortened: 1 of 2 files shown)


php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
```diff
@@ -8,7 +8,13 @@ class PerformanceLevelMinusBaseline implements IGrowthCalculator
 {
     public static function calculateGrowth(StudentCompetency $StudentCompetency)
     {
-        $baseline = $StudentCompetency->BaselineRating;
+        $baseline = round($StudentCompetency->BaselineRating, 1);
+        $performanceLevel = round($StudentCompetency->getDemonstrationsAverage(), 1);
+
+        // baseline & performance level must > 0
+        if (!$performanceLevel || !$baseline) {
+            return false;
```
themightychris, line 16: @nbey shouldn't this be `null`?

php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
```diff
@@ -8,7 +8,13 @@ class PerformanceLevelMinusBaseline implements IGrowthCalculator
 {
     public static function calculateGrowth(StudentCompetency $StudentCompetency)
     {
-        $baseline = $StudentCompetency->BaselineRating;
+        $baseline = round($StudentCompetency->BaselineRating, 1);
+        $performanceLevel = round($StudentCompetency->getDemonstrationsAverage(), 1);
```
themightychris: @nbey let's move the rounding into `getDemonstrationsAverage`, and put a public static config on `StudentCompetency` for the precision, defaulting to `1`. It seems like varying precisions are causing issues for schools and that it would be best for the delivered value to start with consistent rounding.

php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
```diff
@@ -8,7 +8,13 @@ class PerformanceLevelMinusBaseline implements IGrowthCalculator
 {
     public static function calculateGrowth(StudentCompetency $StudentCompetency)
     {
-        $baseline = $StudentCompetency->BaselineRating;
+        $baseline = round($StudentCompetency->BaselineRating, 1);
```
themightychris: @nbey let's create a `getBaselineRating()` getter method and have it apply rounding according to a public static precision config on the `StudentCompetency` class

themightychris (changes requested):

php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
nbey: Currently the method returns `false` when there is no growth to calculate. The `StudentCompetency` class handles returning null, but caches the `false` value.

php-classes/Slate/CBL/Calculators/Growth/PerformanceLevelMinusBaseline.php
nbey: should we look to replace instances of `StudentComptency->BaselineRating` -> `StudentCompetency->getBaselineRating()` in other places as well?

themightychris (approved):

PR #506, opened 2020-01-24
fix: Rating Sheet Issue

nbey:
- [x] feat: disallow closing tabs with ratings


sencha-workspace/packages/slate-cbl/src/field/ratings/StudentCompetenciesField.js
```diff
@@ -290,6 +290,8 @@ Ext.define('Slate.cbl.field.ratings.StudentCompetenciesField', {
             skillIds.push(skillId);
         }
 
+        // disallow closing of cards with atleast one rating
+        card.setClosable(!skillsLength);
```
themightychris: @nbey will this work correctly when there are more than one card in the mix? it looks like the value of `card` is left over from the previous loop, and `skillsLength` won't be scoped to just that card

themightychris (changes requested):

themightychris (approved):

PR #511, opened 2020-02-07
fix: sort skills by code

nbey:
- [x] ensure skills are sorted by `Code` in `SlateDemostrationsStudent` app
- [x] refactor to use `Slate.sorter.Code`

### DIFF
--- a/sencha-workspace/SlateDemonstrationsStudent/app/view/CompetencyCard.js
+++ b/sencha-workspace/SlateDemonstrationsStudent/app/view/CompetencyCard.js
@@ -3,11 +3,13 @@ Ext.define('SlateDemonstrationsStudent.view.CompetencyCard', {
     xtype: 'slate-demonstrations-student-competencycard',
     requires: [
         'Ext.util.Format',
+        'Ext.util.Sortable',
 
         /* global Slate */
         'Slate.cbl.util.Config',
         'Slate.cbl.util.format.PreventOrphans',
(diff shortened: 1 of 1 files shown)


themightychris:
@nbey there's a reusable class for this that's got a bunch of extra smarts we should use. Search the codebase for `Slate.sorter.Code`, it lives in the `slate` repo but is used in a bunch of places in `slate-cbl`. It has a configurable `codeFn` config for extracting code values from items:

```js
        tabPanel.items.setSorters([
            // ...
            new Slate.sorter.Code({
                codeFn: function(item) {
                    return item.selectedCompetency;
                }
            })
        ]);
```

It extends the framework's `Ext.util.Sorter` class that can be dropped into any store. Since there's no store here, I explored a bit how you might use it against an array and found the static [`Ext.util.Sortable.createComparator`](https://docs.sencha.com/extjs/6.2.0/classic/Ext.util.Sortable.html#static-method-createComparator) used like this in `Ext.data.proxy.Memory`:

```js
// ...
Ext.Array.sort(records, Ext.util.Sortable.createComparator(sorters));
// ...
```

It looks like this method can take a single sorter instance or an array

themightychris (changes requested):
use `Slate.sorter.Code`

themightychris (approved):

themightychris:
@nbey in the future, think about how you could incorporate a new Cypress test or assertion into each PR like this

PR #518, opened 2020-03-05
feat: implement archived tasks

nbey:
- [x] implement archived tasks
    - [x] create migration script for archived status
- [x] hide archived tasks by default
- [x] create toolbar option for showing archived tasks in `SlateTasksTeacher`
- [x] update `SlateTasksStudent` filter from `Un-Archived` -> `Archived`
- [x] update cypress tests


themightychris:
@nbey check what happens when you switch sections after enabling "show archived tasks". If you haven't already set it up to include archived tasks when you switch sections, I'd just have the checkbox reset on section switch

nbey:
> @nbey check what happens when you switch sections after enabling "show archived tasks". If you haven't already set it up to include archived tasks when you switch sections, I'd just have the checkbox reset on section switch

@themightychris:

Switching sections with the archived filter will show filtered tasks in the new section as well.

One thing I noticed is that archiving a task when "Show Archived Tasks" is not enabled will not hide the newly archived task in the grid. It does however move it to the bottom, and indicate that its archived. I think this makes sense in this case.

PR #520, opened 2020-03-19
WIP: fix: show orphaned student tasks

nbey:
This PR aims to resolve an issue where student tasks that have a parent task that is not currently being included in the tasks store are not being shown.

- [x] show orphaned student tasks
- [x] sort tasks by due date, title ascending (client-side)
- [ ] redundant server-side sort
- [ ] create cypress test case

### DIFF
--- a/sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
+++ b/sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
@@ -222,11 +222,28 @@ Ext.define('SlateTasksStudent.view.TaskTree', {
             });
         }
 
-        for (; orphanedTasksIndex < orphanedTasksLength; orphanedTasksIndex++) {
-            items.push({
-                studentTask: orphanedTasks.getAt(orphanedTasksIndex).getData()
-            });
-        }
+        items = items.sort(function(taskA, taskB) {
+            taskADue = taskA.studentTask.EffectiveDueDate;
+            taskBDue = taskB.studentTask.EffectiveDueDate;
+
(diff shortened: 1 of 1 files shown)


sencha-workspace/SlateTasksStudent/app/view/TaskTree.js
```diff
@@ -186,33 +186,44 @@ Ext.define('SlateTasksStudent.view.TaskTree', {
             store = me.getStore(),
             items = [],
 
-            rootTasks = store.queryBy(function(studentTask) {
-                return !studentTask.get('Task').ParentTaskID;
+            topLevelTaskIds = [],
+            topLevelTasks = store.queryBy(function(studentTask) {
+                if (!studentTask.get('Task').ParentTaskID) {
+                    topLevelTaskIds.push(studentTask.get('TaskID'));
```
themightychris: @nbey I don't know if it's a good idea to be "doing work" during a query loop, these functions should only return true/false. Since returning true corresponds with this action, I think you could just do something like `topLevelTaskIds = topLevelTasks.collect('TaskID')` on the next line

themightychris (changes requested):

themightychris:
@nbey ping me on slack when this is ready for final review/merge

themightychris:
@nbey did we merge this in another PR already?

nbey:
Nope -- I ran into some issues with getting the server-side sorting to fall-back and replicate the front-end. Also, I think I still need to create a cypress test case for this.

PR #526, opened 2020-04-13
Restore SlateTasksManager application

nbey:
- [x] Restore app functionality
- [x] Remove grid search plugin
- [x] Create basic app smoke test


nbey:
This can be merged once reviewed and all tests pass.

tgaffey:
@nbey I was able to access this app in the sandbox. I successfully created a task with skills and then I was able to clone that task from the task dashboard. The archive button doesnt do anything (which is fine). Are we ready to push this to the skeleton?

nbey:
@tgaffey Yeah, we can push this down to skeleton after it gets merged/released.

Do you think that we should hide the "Archive" button since it hasn't been implemented in this app? It probably makes sense to not show it, since these are all only "public" tasks because they are not tied to a section.

PR #574, opened 2021-02-02
feat(ratings): support arbitrary weights for skill ratings [SCHOOL-24]

nbey:
Closes: SCHOOL-24

- feat: create migrations for new ratings
  - create `EvidenceWeight` column
  - migrate records to use `EvidenceWeight`
  - deprecate `Override` column
- feat: re-implement ability to create `Override` demonstrations
  - chore: remove deprecated field
  - fix: demonstration skill validation logic
  - fix: `SlateDemonstrationsTeacher` app Override creation


php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -252,7 +254,8 @@ public function getDemonstrationData()
                             $demonstrationSkill['DemonstrationDate'] = strtotime($demonstrationSkill['DemonstrationDate']);
                             $demonstrationSkill['TargetLevel'] = intval($demonstrationSkill['TargetLevel']);
                             $demonstrationSkill['DemonstratedLevel'] = intval($demonstrationSkill['DemonstratedLevel']);
-                            $demonstrationSkill['Override'] = $demonstrationSkill['Override'] == '1';
+                            $demonstrationSkill['DemonstrationClass'] = $demonstrationSkill['DemonstrationClass'];
```
themightychris:
```suggestion
```

This line doesn't appear to be needed since there's no casting

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -287,16 +289,29 @@ public function getDemonstrationOpportunities()
...
+                        $this->demonstrationOpportunities += $demonstration['EvidenceWeight'];
                     }
                 }
             }
         }
 
+        if ($hasWildCard) {
+            // return total requirements
+            $this->demonstrationOpportunities = $this->getDemonstrationsRequired();
+        }
```
themightychris: Without this, could `$this->demonstrationOpportunities` ever be greater than `$this->getDemonstrationsRequired();` ? should this block handle capping it too if EvidenceWeight values > 1 are able to produce a higher count than required?

php-classes/Slate/CBL/Demonstrations/DemonstrationSkill.php
```diff
@@ -35,9 +35,10 @@ class DemonstrationSkill extends \VersionedRecord
...
             'default' => null
         ],
-        'Override' => [
-            'type' => 'boolean',
-            'default' => false
+        'EvidenceWeight' => [
+            'type' => 'tinyint',
+            'unsigned' => true,
+            'default' => 1
         ]
```
themightychris, line 43: - [ ] double-check the current codebase for any other instances of `Override` that might need to be removed/ported

php-classes/Slate/CBL/StudentCompetency.php
themightychris: This doesn't look like it was previously working at all due to case mismatch in `$hasWildcard` vs `$hasWildCard`

themightychris: fixed in d2b35a077f6be09cc64f996ee7ad97c326cc2093

PR #576, opened 2021-02-16
Docs: Student Competency Methods [SCHOOL-57]

nbey:
Documentation for methods defined in `Slate\CBL\StudentCompetency`


themightychris (changes requested):
@nbey content looks good, let's merge this as it is before you update the content to the new rules, but please run it through a markdown linter first (install the vscode extension)

PR #585, opened 2021-05-18
[SCHOOL-45] Update student taskfilter behavior

BillClinton:
changed the behavior of the "View All" button to check all statuses
including archived per SCHOOL-45


themightychris (changes requested):
@BillClinton can you make it dismiss the menu after you click view all too?

PR #586, opened 2021-06-19
Scaffold Vue app for new advanced portfolio manager [SCHOOL-89]

ryon:
## What's here

- Trying out a setup for a new Vue 2 workflow, with Vue CLI
- Mocking up rough layout for Advanced Portfolio Manager

## Open questions

- Is Vue CLI's NPM setup (uses Webpack... somewhere in there) the right choice? Or something like Parcel, Vite? Or more custom?
- Is this ESLint config good?
- Bootstrap and BootstrapVue bring a lot of value and developer convenience previously afforded by the Emergence Skeleton frontend and Ext JS. Are they the right choices? Do we want to exclude both of the latter in the new setup?
- Should we get ahead of Vue 3 instead of using the more stable/compatible version 2?
- FontAwesome's Vue package is a little verbose/annoying to use (each individual icon must be explicitly imported, and the component syntax is a lot of typing). Alternate approach?

Screen Shot 2021-06-19 at 4 47 56 PM


ryon:
Update: flesh out the enrollments grid a little, including row/column highlights and passing skill/student data between the two components

themightychris (approved):

chriscauley:
#### Is Vue CLI's NPM setup (uses Webpack... somewhere in there) the right choice? Or something like Parcel, Vite? Or more custom?

I use vue-cli for all my projects. Vite is fine, but I stopped using it because it gave me issues. I think vue-cli is more stable. Either will give you access to the broader vue ecosystem (eg you can just run `vue add tailwind` which is a pain to setup otherwise), so I'd advise against using anything else. Using parcel or rolling your own webpack is going to be a lot of work.

----

#### Is this ESLint config good?

I was frustrated that it didn't let me use `optional?.chaining` or the `nullish ?? operator`. I make a recommendation to update to vue-3 below, which would give you a better set of defaults. If you want to get deep into my personal preferences, all my projects use [@unrest/eslint-config](https://github.com/chriscauley/unrest-eslint-config/blob/main/index.js). Jarvus should probably have something like this.

----

#### Bootstrap and BootstrapVue [...]

The only time I recommend packages like this are for personal projects. You have a designer and that means that you're going to write more CSS undoing bootstraps styles than you save by using this. Also bootstraps utility classes use `!important` which in my opinion makes them unusable. I could put together a 10 minute presentation on how to use tailwind in a vue project.

The problem is that (unless you're on a personal project where you can just accept the framework for what it is), you'll have to spend a ton off effort hacking in new styles and behaviors that the client wants. These hacks make future changes much more difficult. VueBootstap also results in a lot of components like `<b-col>` which literally only add a single class to a div.

----

#### Vue 2 vs 3

My preference is Vue 3. 
* I had trouble finding tutorials for vue2 for some of the packages I was using. The community has been slow to adopt but it is shifting.
* Having to call `Vue.set(myDict, "myKey", myValue)` is annoying. It can also be a gotcha if people forget and then don't thoroughly test.
* Supposedly you can use Pinia stores outside of components, but it didn't work for me. Upgrading to vue 3 will fix this.
* Optional chaining and nullish operators didn't work in the vue2 templates. It's a small nitpick, but it really is nice to type `this.userStore.get()?.data` instead of 
```
const response = this.userStore.get()
return response && response.data
```

----

#### FontAwesome's Vue package [...] Alternative approach?

I prefer just to use the font style sheet. The svg icons will eventually take up more space if you use a lot of them and having to import them all is annoying. I've never understood what the `<font-awesome-icon>` tag gets you. My recommendation:
* In public.html: `<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.css" />
* In any template: `<i class="fa fa-edit" />`

PR #593, opened 2021-08-18
Working on disabled test

Camus1859:


Camus1859:
In student_task.js im unable to fix the bug on line 341

cypress/integration/student_demonstrations_growth.js
```diff
@@ -37,24 +37,25 @@ describe.skip('Student competency growth calculation test', () => {
...
                             const studentContentAreas = Object.keys(growthCalculationsByStudent[studentUsername]);
                             studentContentAreas.forEach(studentContentArea => {
                                 cy.visit(`/cbl/dashboards/demonstrations/student#${studentUsername}/${studentContentArea}`);
+                                cy.request(`/cbl/dashboards/demonstrations/student#${studentUsername}/${studentContentArea}`).then((res) => {
+                                  expect(res.status).to.eq(200)
+                                })
                                 // ensure that API has loaded required data
                                 cy.wait('@studentCompetencyData')
-                                    .its('status')
-                                    .should('be', 200)
```
nbey: I suggest we revert this part back to the original code, with one minor change: 
*Line 49*
Replace: `.should('be', 200)`
With: `.should('eq', 200)`

cypress/integration/login_admin.js
```diff
@@ -10,15 +10,15 @@ describe.skip('Admin login test', () => {
 
         // should be visible after Log In click
         cy.get('#login-modal').should('not.be.visible');
-        cy.contains('Log In').click();
+        cy.get('.mobile-hidden').contains('Log In').click();
```
themightychris: It's generally a bad idea to rely on incidental selectors like `.mobile-hidden` as part of cypress testing.

I'm also wondering why this became necessary, did you find that the width of the screen in tests was putting the design in mobile layout and hiding the login button? Did this extra selector make it work even though it was hidden? I wouldn't expect adding this before the same `.contains('Log In')` to fix anything as AFAICT it would just further restrict the contains check?

nbey: @themightychris 

https://github.com/SlateFoundation/slate/commit/98c4de86319284165d10856af3b28e91823469c1#diff-fc3befdd534c43f9355501bd6768ec67a8fcab60563771e0453a06134c35d37a

I _believe_ this commit changed the template so there's two sections, and without specifying `cy.get` is grabbing the first, which happens to be the mobile-only (hidden) Log In.

cypress/integration/student_demonstrations_growth.js
```diff
@@ -37,24 +37,25 @@ describe.skip('Student competency growth calculation test', () => {
                 };
 
                 cy.visit('/cbl/dashboards/demonstrations/student').then(() => {
-                    cy.withExt().then(({extQuerySelector}) => {
+                    cy.withExt().then((res) => {
```
themightychris: why create this `res` variable here instead of leaving the previous code that just pulled `extQuerySelector` out?

I see below we replaced `extQuerySelector` with `res. extQuerySelector`. The only other use of `res` is when it's used in `cy.request`'s callback. In that case, a new variable called `res` is "shadowing" over the different variable with the same name in a higher function scope. This is considered a very bad practice in javascript (re-using the same variable name in a lower scope) -- if you had your `eslint` extension in vscode working correctly it should have warned you about that -- make sure your eslint is working and showing  a warning here. It makes it very easy to make mistakes when there are two different variables with the same name overlapping like this.

You should change the outer use of `res` back to how it was, it will have no impact on the inner use of res. Then, avoid using variable names like `res` that are abbreviations. It's best to use full words in variable names and be explicit "what" think you're talking about.  Consider a name like `dashboardResponse` instead

nbey (changes requested):
Looks like the  `unionfs-fuse` submodule was accidentally checked in -- let's remove that before merging.

cypress/integration/student_demonstrations_growth.js
```diff
@@ -37,24 +37,25 @@ describe.skip('Student competency growth calculation test', () => {
...
                         const studentUsernames = Object.keys(growthCalculationsByStudent);
                         studentUsernames.forEach(studentUsername => {
                             const studentContentAreas = Object.keys(growthCalculationsByStudent[studentUsername]);
                             studentContentAreas.forEach(studentContentArea => {
                                 cy.visit(`/cbl/dashboards/demonstrations/student#${studentUsername}/${studentContentArea}`);
+                                cy.request(`/cbl/dashboards/demonstrations/student#${studentUsername}/${studentContentArea}`).then((res) => {
+                                  expect(res.status).to.eq(200)
+                                })
```
themightychris: why did we add this `cy.request` call? You would normally use `cy.request` to make API calls. In this case we appear to be loading the same page we just navigated the browser to a second time via XHR. The `cy.visit` call would already have failed if the browser was unable to navigate to the page.

I would remove this `cy.request` call -- was there any particular reason you added it?

cypress/integration/student_demonstrations_growth.js
themightychris: This code was doing something different than what maybe the `cy.request` you added was aiming to do. `cy.wait` waits for the application loaded into the page to make an API call on its own, and lets you check the status of that. This accomplishes two things for us here: has the test check that an API call we expect the page to make succeeded, but also serves to make the test script wait for the API call to finish before the next assertion (otherwise the next assertion might time out if the API call is slow some time)

cypress/integration/teacher_demonstrations.js
```diff
@@ -1,84 +1,83 @@
-describe.skip('Teacher demonstrations test', () => {
```
themightychris, line 1: it looks like you reindented this whole file to 2-space indent from 4-space indent.

I'm not against the idea of changing the indentation standard for cypress tests from 4-spaces to 2-spaces if that's more standard in the community, but if we were to do that we should do it across all cypress tests at the same time, and do it in its own PR not mixed in with other changes. Please revert this conversion from this PR so that the PR only shows content changes to the test without changes to spacing mixed in

script/test
```diff
@@ -46,3 +46,7 @@ echo "==> test: setting up node_modules…"
 echo
 echo "==> test: executing \`cypress run\`…"
 (cd "${temp_path}" && npx cypress run)
+
+
+
+
```
themightychris: Always review all your changes as you stage them for commit, and again under the Files Changed view once you open your PR, and try to minimize extraneous changes

themightychris (changes requested):
looking good, but some PR-hygiene work is needed

I couldn't comment inline, but we want to remove the `unionfs-fuse` object from this PR too. It looks like you cloned that repo into your slate-cbl working tree and then committed a pointer to its commit

cypress/integration/student_demonstrations_growth.js
Camus1859: Got it

cypress/integration/student_demonstrations_growth.js
Camus1859: I added `cy.request` because i saw that 
`cy.wait('@studentCompetencyData')
                                  .its('status')
                                  .should('eq', 200)
                                  .then(() => {`
                                  
  wasnt returning the correct status. When I looked at the docs one of the first things that came up was showing how to check that `status` of the `req` object which I thought we were doing. So I figured I would use `cy.request(...)`  and then use the `res` object to check the status for 200 and then the code passed so I figured my solution was correct.

cypress/integration/teacher_demonstrations.js
Camus1859: My prettier extension tab indent was on 2. I changed it to 4. Its weird that it only did this to this specific file and not the others..

nbey:
This is an older version of #603

PR #598, opened 2021-08-24
Update formatting tools

Camus1859:


themightychris:
@Camus1859 this PR has sat open for a while, which is really dangerous for a bulk reformat PR. Let's do this to close this out:

1. Start a fresh branch on the latest develop, and make fresh commits adding the config files you've developed
2. Document in the PR via a `script/*` bash script command or an article under `docs/development` a command we can run that bulk delints all the test scripts according to the formatting config added in step 1
3. _do not_ commit any formatting changes to this PR until it is reviewed, but test out the command and review the changes it makes before discarding them

The reviewer for the PR will test the reformatting command locally and review that, and if it looks good they'll handle committing the reformatting and then merging the PR immediately so we avoid any gap of time where formatting changes sit out and risk reverting real work from other branches

PR #600, opened 2021-08-26
refactor: student_demonstrations_growth test [SCHOOL-98]

nbey:
refactor: student_demonstrations_growth test
- deprecate json fixtures file


cypress/integration/student_demonstrations_growth.js
```diff
@@ -252,5 +252,316 @@ describe('Student competency growth calculation test', () => {
...
+        checkCompetencyCalculation({
+            student: 'student4',
+            contentArea: 'ELA',
+            competency: 6,
+            baseline: NULL,
+            growth: NULL
+        });
+    });
+    
+    it.skip('One full ER plus one additional rating, () => {
```
nbey: Looks like you missed the closing `'` tag.

cypress/integration/student_demonstrations_growth.js
```diff
@@ -252,5 +252,316 @@ describe('Student competency growth calculation test', () => {
...
+        checkCompetencyCalculation({
+            student: 'student4',
+            contentArea: 'ELA',
+            competency: 7,
+            baseline: 6,
+            growth: 0.4
+        });
+    });
+    
+    it.skip('One full ER plus one additional rating, less than 50%    ', () => {
```
nbey: Looks like some spaces were pasted at the end of the description

themightychris (changes requested):
So I've realized that the reason we _need_ the JSON fixtures file is because the `b21-skeleton` repo (and potentially others) have to be able to [override it with different values/cases](https://github.com/b21network/b21-skeleton/tree/develop/cypress/fixtures) because subprojects can potentially change calculations (and b21-skeleton DOES by setting a different growth calculation method)

In #617 I went the other direction to make heavier use of the fixture file, and I think we need to do the same thing here for these tests to pass downstream on `b21-skeleton` where the growth numbers and math have to be different

nbey:
@themightychris @Camus1859 This PR can be closed -- the original intent for this PR was to deprecate the fixtures file, but we now know that is no longer the route we want to go. There were some minor updates made to the test logic, however I believe the original fixtures file / test should suffice.

PR #603, opened 2021-09-07
Enable disabled test [SCHOOL-99]

Camus1859:


.editorconfig
```diff
@@ -0,0 +1,19 @@
+root = true
```
nbey: Let's move this file into it's own PR (I think you have one created already)

themightychris (approved):

nbey (approved):

themightychris:
This PR seems good to merge aside from whatever 500 crash is happening

themightychris (approved):

cypress/integration/login_admin.js
```diff
@@ -10,15 +10,15 @@ describe.skip('Admin login test', () => {
 
         // should be visible after Log In click
         cy.get('#login-modal').should('not.be.visible');
-        cy.contains('Log In').click();
+        cy.get('.slate-omnibar.mobile-hidden').contains('Log In').click();
```
themightychris, line 13: @Camus1859 in the future, it's better to augment the source application with semantic `data-*` attributes for tests to anchor to than to rely on any presentation-centric CSS tags

PR #605, opened 2021-09-22
refactor: student tasks tests [SCHOOL-109]

nbey:


cypress/integration/student_tasks.js
```diff
@@ -205,140 +92,515 @@ describe.skip('Student tasks test', () => {
...
+                    .find('ul.slate-tasktree-list')
+                    .children('li.slate-tasktree-item')
+                    .each(($item) => {
+                        cy.get($item).find('.slate-tasktree-status').contains('Completed'); // confirm the status text exists
+                    });
+            })
+        })
+    })
+
+    //this test could be shorter if we make an API call and test the data instead of clicking on every li element
```
Camus1859: These comments refer to 6 out of 7 of my commits.


1) Im clicking on every Li element in order to get the full date on the modal that appears (this obviously is causing the test to take much longer than needed to run, we could possible make an API call here and check this) in this format `YYYY-MM-DD`. Im not sure how you were able to use the  date variable that appears on the Li element. I had issues doing this because the date variable was not in the correct format to get passed into a dayjs method.
Im not sure how this works because date should be in this format `YYYY-MM-DD` and I dont think it is  `...isAfter(date, 'day')...`

```let today = dayjs(),
       date = dayjs($date.text()),
       isAfter = today.isAfter(date, 'day') || today.isSame(date, 'day');
```

2) I used the CSS selector to get the modal body. I know this is not maintainable and we will need to use Sencha to get the component but I wasnt sure what the Sencha component was and I figured I would use the CSS selector for now just to get the test completed and we would be able to easily add the Sencha component

3) Im waiting for 1 second for the modal to load and 1 second for it to be removed from the DOM.

themightychris:
@nbey this PR needs a rebase but hit some minor conflicts so I skipped it for now, plz handle when you can

nbey:
Rebased and re-ran the tests. This PR should be good to go now.

holly-g:
Awesome, thanks @nbey !

themightychris (approved):

PR #613, opened 2021-10-08
refactor: added API, CSV, and UI test [SCHOOL-115]

Camus1859:


cypress/integration/precise-rounding.js
```diff
@@ -0,0 +1,395 @@
+const csvtojson = require('csvtojson');
+const dayjs = require('dayjs');
+const isBetween = require('dayjs/plugin/isBetween');
+
+dayjs.extend(isBetween);
+
+const testCaseOne = {
```
themightychris: **suggestion (non-blocking):** It might be good to stash all these in a fixture file, see [these examples](https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/fundamentals__fixtures) and [these docs](https://docs.cypress.io/api/commands/fixture)

cypress/integration/precise-rounding.js
```diff
@@ -0,0 +1,395 @@
...
+
+            expect(expectedProgress,
+                `${contentArea}.${competency} for ${student} CSV Completion Percentage Value ${expectedProgress}: Test Doc Completion Percentage Value ${progress}`
+                ).to.equal(progress/100); // progress is represented as decimal in export
+        });
+        });
+    };
+
+
+  it('Student 5 ELA Competency 7', () => {
```
themightychris: **suggestion (non-blocking):** We generally try not to get too engineer-y with tests, but this might be a good case for generating all these `it()` blocks within a loop through all the test cases loaded from a fixture file as I mentioned above. You might just put all the statements as a property inside there, like:

```json
[
  {
    "caseTitle": "Student 5 ELA Competency 7",
    "student": "student4",
    "contentArea": "ELA",
    "competency": "7",
    "baseline": "6",
    "growth": "0.4",
    "progress": "63",
    "performanceLevel": "6.4"
  },
  {
    // ...
  }
]
```

cypress/integration/precise-rounding.js
```diff
@@ -0,0 +1,395 @@
...
+    baseline: null,
+    growth: null,
+    progress: "0",
+    performanceLevel: "-"
+};
+
+describe('Confirm rounding is consistent across UI, API, and exports', () => {
+
+    // API Tests
+    const checkAPIDataAgainstDocsTestData = ({student, contentArea, competency, baseline, growth, progress, performanceLevel}) => {
```
themightychris: **suggestion (non-blocking)**: it should be possible to just declare these as functions at the bottom of this file, outside the describe, e.g.:

```javascript

describe('Confirm rounding is consistent across UI, API, and exports', () => {
  // ...
});

function checkAPIDataAgainstDocsTestData = ({student, contentArea, competency, baseline, growth, progress, performanceLevel}) {

}
```

Such functions will get "hoisted" to existing throughout the file even though they're at the bottom. This will make the `describe` block more readable by reducing the testing

**suggestion (non-blocking)**: I'd remove `DocsTestData` as future maintainers won't be aware of the doc we drew test cases out of and it will confusing extra terminology. So this might just be `checkAPIDataAgainstTestCase`

themightychris (approved):

cypress/integration/precise-rounding.js
themightychris: @Camus1859 also note how I quoted the numbers as stings. JavaScript has a lot of quirks when it comes to its `Float` type which is what you get whenever you have an unquoted number with a dot in it. Basically JavaScript has no way to handle fixed-precision numbers natively, and the sorts of weird stuff that happens is along the lines of what we're likely to run into comparing fixed precision numbers of different roundings. Our best bet is to keep our test case data as strings, capture actuals without ever converting them away from strings (in the case of UI and CSV) or turn them into strings immediately and never touch them with math operators (in the case of JSON responses), and do strict equality checks on strings vs strings so JavaScript is never tempted to coerse types

cypress/integration/precise-rounding.js
```diff
@@ -0,0 +1,191 @@
...
+            expect(baseline,
+                `${contentArea}.${competency} for ${student} CSV Baseline Value ${csvBaseLine}: Fixtures data Baseline Value ${baseline}`
+            ).to.be.oneOf(['0', null])
+        } else {
+            expect(`${csvBaseLine}`,
+                `${contentArea}.${competency} for ${student} CSV Baseline Value ${csvBaseLine}: Fixtures data Baseline Value ${baseline}`
+            ).to.equal(baseline);
+        }
+
+        // csv represents 0 progress as empty string
```
themightychris: I _think_ the spreadsheets should have 0, not blank, for progress values where the value is 0. The only case where blank would make sense is if progress in null because a given competency has 0 demonstration requirements for whatever reason. Otherwise it should also have a numeric value and we should render 0's where a calculation could be made and resulted in 0

Is that right @tgaffey ? would we break any of your processes to start having 0s instead of blanks in progress values?

php-classes/Slate/CBL/StudentCompetency.php
```diff
@@ -155,6 +155,7 @@ public function getCompletion()
             'CompetencyID' => $this->CompetencyID,
             'currentLevel' => $this->Level,
             'baselineRating' => $this->BaselineRating,
+            'baselineRating2' => $this->getBaselineRating(),
```
themightychris: @Camus1859 can we delete this line before merging? or even better: delete/edit the commit that added it

themightychris: This change is still showing under GitHub's Files Changed tab so its still in the PR

cypress/integration/precise-rounding.js
```diff
@@ -0,0 +1,218 @@
...
+                    .then(() => {
+                        cy.wait(500); // wait for dom to render
+
+                        // ensure competency card elements have rendered
+                        const studentCompetencySuffixes = Object.keys(testCases[studentUsername][studentContentArea]);
+                        studentCompetencySuffixes.forEach(studentCompetencySuffix => {
+                            cy.get('li.slate-demonstrations-student-competencycard')
+                                .then(() => {
+                                    cy.withExt().then(({extQuerySelector}) => {
+                                        const card = extQuerySelector(`slate-demonstrations-student-competencycard{getCompetency().get("Code")=="${studentContentArea}.${studentCompetencySuffix}"}`);
```
themightychris, line 85: @Camus1859 forget what I said on the call about not using selectors with generated IDs, this code is resolving those dynamically based on stable selectors

nbey:
Looks like all tests are passing, except for blog tests. I've created a PR that should hopefully address those issues here: https://github.com/JarvusInnovations/emergence-skeleton/pull/309 otherwise, I think it looks good to merge.

nbey (approved):

PR #617, opened 2021-10-13
fix(e2e): update exports/student_competencies test to use calculations fixture

themightychris:
These tests need to be able to use different numbers when they run on b21-skeleton because it uses a different growth model calculator. We already had such a file with the values we need in it in place on `slabe-cbl` and `b21-skeleton` as part of other tests. I renamed that file a bit in this PR to reflect its broadening use to cover all calculation value test cases

I looked at this example, among its siblings and those for fixtures, to get this working: https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/fundamentals__dynamic-tests/cypress/integration/request-spec.js

The key findings there was that the generation of tests:

- could not be driven by anything async. There can be no async operation between the `describe` body executing and it executing calls to `it()` to define tests
- `require(...)` can be used to load JSON fixtures synchronously and therefor be able to use fixture data to generate `it()` calls
- anything async that needs to happen ahead of all the tests can be put in `before` and CAN share data with test runs via mutable variables within the shared scope


themightychris:
@nbey @Camus1859 I'm going to go ahead and merge this because I want to see if it gets b21-skeleton passing its tests upstream, but I would appreciate your reviews or questions retroactively

PR #632, opened 2021-12-07
fix: completion percentage on skills with no ER [SCHOOL-56]

nbey:


themightychris (approved):
great work! and good catch on the `getTotalDemonstrationsRequired` method

sencha-workspace/SlateDemonstrationsTeacher/app/view/ProgressGrid.js
```diff
@@ -769,10 +769,10 @@ Ext.define('SlateDemonstrationsTeacher.view.ProgressGrid', {
             countDirty = count != node.renderedCount;
             averageDirty = average != node.renderedAverage;
             levelDirty = level != renderedLevel;
-            demonstrationsRequired = competency.totalDemonstrationsRequired[level] || competency.totalDemonstrationsRequired.default;
+            demonstrationsRequired = competency.getTotalDemonstrationsRequired(level);
```
themightychris, line 772: @nbey FYI, I have no idea how the E2E passed on this PR, but `competency` doesn't have `getTotalDemonstrationsRequired` because in this spot it's just an embedded data object, not a model instance. I'm patching that now but keep an eye out for an explanation of how this change could have been in this PR yet the E2E passed -- they've failed in every branch after this

PR #636, opened 2021-12-31
Show non enrollment in student demonstrations dash [SCHOOL-137]

BillClinton:


sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
```diff
@@ -232,7 +243,8 @@ Ext.define('SlateDemonstrationsStudent.view.Dashboard', {
             me.getCompetenciesSummary(),
             me.getRecentProgress(),
             me.getLegend(),
-            me.getCardsCt()
+            me.getCardsCt(),
+            me.getNonEnrollmentMessage()
```
themightychris, line 274: This isn't working how the others are, they all have a transformation in place that turn the config into an instance with some sugar that translates true/false into hidden: false/true. Replicate that for this new "slot component" for consistency and then we can do `me.setNonEnrollmentMessage(!studentEnrolled)`

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
```diff
@@ -98,6 +98,17 @@ Ext.define('SlateDemonstrationsStudent.view.Dashboard', {
          */
         legend: false,
 
+        /**
+         * @cfg {Ext.container.Container|Object|boolean}
+         * Instance or configuration for container containing non-enrollment message.
+         */
+        nonEnrollmentMessage: {
+            xtype: 'container',
```
themightychris: don't use `container` where `component` will do

sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
```diff
@@ -22,7 +22,10 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
         studentSelector: 'slate-demonstrations-student-dashboard slate-appheader slate-cbl-studentselector',
         contentAreaSelector: 'slate-demonstrations-student-dashboard slate-appheader slate-cbl-contentareaselector',
         competenciesSummary: 'slate-demonstrations-student-competenciessummary',
-        cardsCt: 'slate-demonstrations-student-cardsct'
+        recentProgress: 'slate-demonstrations-student-recentprogress',
+        levelsLegend: 'slate-cbl-levelslegend',
+        cardsCt: 'slate-demonstrations-student-cardsct',
+        nonEnrollmentMessage: 'slate-demonstrations-student-dashboard container[itemId="non_enrollment_message"]'
```
themightychris: FYI, this is the preferred way to utilize an itemId in a selector:

```suggestion
        nonEnrollmentMessage: 'slate-demonstrations-student-dashboard #non_enrollment_message'
```

But see other feedback, none of these selectors are needed

sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
```diff
@@ -250,6 +256,9 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
 
         Ext.resumeLayouts(true);
 
+        // show/hide components based on enrollment
+        [competenciesSummary, recentProgress, levelsLegend, cardsCt ].forEach(cntr => cntr.setVisible(studentEnrolled));
```
themightychris: For all these components, we use a "subcomponent slot" strategy where the config system can be used to tell the main component which subcomponents should be hidden/visible without knowing the details of how hiding/showing them should work. We can get rid of all the added refs, and instead of getting a ref to each subcomponent we just want to do `mainComponent.setSubComponent(shouldBeVisible)`

themightychris (changes requested):
Looks good, but I provided some stylistic/architectural feedback to help conform the approaches here to the architecture we use for managing interacting with subcomponents. I think implementing these patterns here should be quick, will help keep consistency, and is good practice for the pattern we want to use going forward

sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
BillClinton: @themightychris I implemented this here:  https://github.com/SlateFoundation/slate-cbl/commit/a5c33a6c3832f6061ee0a9511bf9502838bee16c

sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
```diff
@@ -22,7 +22,10 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
         studentSelector: 'slate-demonstrations-student-dashboard slate-appheader slate-cbl-studentselector',
         contentAreaSelector: 'slate-demonstrations-student-dashboard slate-appheader slate-cbl-contentareaselector',
         competenciesSummary: 'slate-demonstrations-student-competenciessummary',
-        cardsCt: 'slate-demonstrations-student-cardsct'
+        recentProgress: 'slate-demonstrations-student-recentprogress',
+        levelsLegend: 'slate-cbl-levelslegend',
+        cardsCt: 'slate-demonstrations-student-cardsct',
+        nonEnrollmentMessage: 'slate-demonstrations-student-dashboard #non_enrollment_message'
```
themightychris: @BillClinton these changes can be removed now, we're accessing these subcomponents entirely through their config slots rather than via controller refs

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
```diff
@@ -221,6 +246,16 @@ Ext.define('SlateDemonstrationsStudent.view.Dashboard', {
...
     },
 
+    applyNonEnrollmentMessage: function(message, oldMessage) {
+        if (typeof message === 'boolean') {
+            message = {
+                hidden: !message
+            };
+        }
+
+        return Ext.factory(message, 'SlateDemonstrationsStudent.view.NonEnrollmentMessage', oldMessage);
```
themightychris, line 259: @BillClinton this PR doesn't appear to be adding a new file for this new class `SlateDemonstrationsStudent.view.NonEnrollmentMessage` so this won't work

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
```diff
@@ -98,6 +98,18 @@ Ext.define('SlateDemonstrationsStudent.view.Dashboard', {
...
+         * @cfg {Ext.Component|Object|boolean}
+         * Instance or configuration for container containing non-enrollment message.
+         */
+        nonEnrollmentMessage: {
+            xtype: 'component',
+            itemId: 'non_enrollment_message',
+            cls: 'slate-placeholder',
+            html: 'Not currently enrolled in this competency',
+            hidden: true
+        },
```
themightychris: If we're using a factory to build this instance and creating the `NonEnrollmentMessage` view class as referenced in the comment lower down, all this content should become part of that class and this config should just be defaulting to `false` (hidden) like the others

Also if we're moving this to a dedicated component with its own class name/xtype, it won't need `itemId` anymore

themightychris (changes requested):

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
BillClinton: This just worked, maybe because the second 2 arguments are optional.   But I should have looked into the factory method anyway and just sent the first argument if it is working that way

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
BillClinton: @themightychris It felt a bit clunky to create a component class in its own file to display something that's similar to a "data not found" error.  But using an Ext.window.MessageBox seemed like overkill too.  I considered changing the text of the placeholderItem, but that felt hacky.  But it does make the Dashboard view cleaner: https://github.com/SlateFoundation/slate-cbl/pull/636/commits/bdb91989a90b2528573cbdd8eaf95bb778efced5  If there's a more Slate standard way to do this let me know.

themightychris: @BillClinton yeah great instinct. Instead of creating a new class, we could have also just had the `apply*` method fill in config for a base `Ext.Component` factory like this example does for a `TextArea`: https://github.com/SlateFoundation/slate-cbl/blob/develop/sencha-workspace/packages/slate-cbl/src/field/comments/Field.js#L50

But since you had already written up a name for the new class in the code I decided not to overcomplicate my feedback. I probably would have went this route instead of making the class in this case. The nice thing about having the class though is that sometimes we create a school-specific Sencha package that is just a bundle of class overrides, and having things broken out into named classes with copy/templates in config defaults makes it really easy for overrides to customize things

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
```diff
@@ -149,11 +156,24 @@ Ext.define('SlateDemonstrationsStudent.view.Dashboard', {
         me.setRecentProgress(contentAreaSet);
         me.setLegend(contentAreaSet);
         me.setCardsCt(contentAreaSet);
+        me.setNonEnrollmentMessage(contentAreaSet);
         Ext.resumeLayouts(true);
 
         me.fireEvent('selectedcontentareachange', me, contentArea, oldContentArea);
     },
 
+    updateCardsForEnrollmentStatus: function(enrolled) {
```
themightychris: @BillClinton so, final nitpick. `update*` and `apply*` function names on components are used by the config system and should be avoided for methods that aren't config lifecycles.

That said, I was thinking this earlier, this is actually a good case for using the config system to mediate state changes. Put `hasEnrollments: true` up above and then rename this to `updateHasEnrollments` to match, and then the call from the outside becomes `component.setHasEnrollments(trueOrFalse)` -- the benefit there then is that the config system handles only calling `update*` when the value changes. So calling `component.setHasEnrollments(true)` after the last call was `component.setHasEnrollments(true)` won't run the `update` method a second time. It's a nice little mechanic that lets you "assert" the desired state aggressively and then let the config system handle debouncing and only running the update code when there's a change

If making a config didn't make sense, I'd typically prefix a method like this with `sync*`

themightychris (changes requested):
one tweak and then we're good to merge!

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
BillClinton: 👍

themightychris (approved):

PR #638, opened 2022-01-03
Fix/school 124 use competency calculation json on UI api csv test

Camus1859:


cypress/integration/student-competency-calculations.js
```diff
@@ -21,8 +21,8 @@ describe('Confirm rounding is consistent across UI, API, and exports', () => {
                     .should(({ xhr }) => {
                         const studentCompetencySuffixes = Object.keys(testCases[studentUsername][studentContentArea]);
```
themightychris: **nitpick** lets rename `studentCompetencySuffixes` since what it contains is no longer suffixes, they are now `studentCompetencyCodes`

It makes the rest of the code confusing to read

themightychris (changes requested):

cypress/integration/student-competency-calculations.js
Camus1859: done

PR #639, opened 2022-01-20
Prevent parent tasks from being made subtasks [SCHOOL-146]

BillClinton:
The tasks UI currently is only able to handle one level of parent->child task nesting, but the task edit form doesn’t stop you from assigning parent tasks across multiple levels. The result right now of doing this is that child tasks past the 2nd level disappear from the UI and aren’t reachable

This PR adds frontend and backend restrictions on selecting a parent task for tasks that already have children.  On the backend, attempting to save a parent task that has a parentTaskID will cause an error.  On the frontend, the parent task field will be hidden when editing parent tasks

### DIFF
--- a/sencha-workspace/packages/slate-cbl/src/view/tasks/TaskForm.js
+++ b/sencha-workspace/packages/slate-cbl/src/view/tasks/TaskForm.js
@@ -254,7 +254,7 @@ Ext.define('Slate.cbl.view.tasks.TaskForm', function() {
                 );
 
                 // Hide the parentTask field if this task is a parent of other tasks
-                me.getParentTaskField().setHidden(task.get('ChildTasks').length > 0);
+                me.getParentTaskField().setHidden(task.get('SubTasks').length > 0);
 
                 me.getClonedTaskField().setHidden(!task.phantom);
                 me.getClonedTaskDisplayField().setHidden(task.phantom);
--- a/sencha-workspace/packages/slate-cbl/src/view/tasks/TaskForm.js
+++ b/sencha-workspace/packages/slate-cbl/src/view/tasks/TaskForm.js
@@ -253,6 +253,9 @@ Ext.define('Slate.cbl.view.tasks.TaskForm', function() {
                         : Ext.String.format(me.getInitialConfig('editTitle'), task.getId(), task.get('Title'))
                 );
 
+                // Hide the parentTask field if this task is a parent of other tasks
+                me.getParentTaskField().setHidden(task.get('ChildTasks').length > 0);
+
                 me.getClonedTaskField().setHidden(!task.phantom);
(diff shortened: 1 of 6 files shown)


themightychris (approved):
well done

themightychris:
@BillClinton this doesn't seem to be working in a couple cases:

## Case 1

Logged in as `teacher` / `teacher` here: https://pr-639.slate-cbl.sandbox.k8s.jarv.us/cbl/dashboards/tasks/teacher#ELA-002/all

If I edit the milestone 6 task, the "subtask of" dropdown is still visible despite this task already having child tasks:
Screen Shot 2022-01-20 at 6 45 11 PM

It does however correctly error when I try to save

## Case 2

While creating a new task, if I select an existing task like "Milestone 3" that is a top-level task with no children, I get the parent task error even though this hsould be ok: 
Screen Shot 2022-01-20 at 6 47 32 PM

themightychris (changes requested):
see error cases

sencha-workspace/packages/slate-cbl/src/view/tasks/TaskForm.js
```diff
@@ -253,6 +253,9 @@ Ext.define('Slate.cbl.view.tasks.TaskForm', function() {
                         : Ext.String.format(me.getInitialConfig('editTitle'), task.getId(), task.get('Title'))
                 );
 
+                // Hide the parentTask field if this task is a parent of other tasks
+                me.getParentTaskField().setHidden(task.get('ChildTasks').length > 0);
```
themightychris: @BillClinton it looks like the reason this isn't working is because ChildTasks is dynamically populated in the main view while a fresh task instance is loaded from the server when opening the modal in order to grab some additional embedded data and that load doesn't include the dynamic field `SubTasks` or the populated `ChildTasks` that the main store has.

The reason `ChildTasks` is populated client side rather than using the server-side `SubTasks` is because arbitrary filters can be applied to the set of tasks loaded into the main view, and we want to only show the child tasks that are part of that filtered set rather than all subtasks of any task that is part of that filtered set.

There is a bit of an issue here though in that the dynamically-populated local field is called `ChildTasks` while the matching dynamic field is called `SubTasks`. We should probably replace `ChildTasks` every where that exact case-sensitive string is used with `SubTasks` to match the server (even though it can be populated locally with slightly different logic, it is still the same relationship)

Then add `SubTasks` to the default `include` list in the `Task.js` model -- this is the standard set of subfields to include when loading an individual task. The `Tasks` store declares its own proxy to bypass using the Model's proxy so a more no dynamic fields get embedded when loading lists of tasks in the main view, so we can covert `ChildTasks` to `SubTasks` and then load `SubTasks` in the model include spec without conflicting with how the main store loads the Tasks list and then internally populates sub tasks

themightychris (approved):
well done!

PR #647, opened 2022-02-08
Powertools menu missing the power tools [SCHOOL-157] 

BillClinton:
CBL powertools were not showing up in the new powertools menu because they were in the cbl/powertools directory rather than a powertools/cbl directory.  This PR moves the powertools to the powertools/cbl directory and creates an _index.php for the new directory that calls the index request handler.

@themightychris There are a number of files in the site-root/cbl directory.  I only moved the files from site-root/cbl/powertools.  Could you take a look at the files in the site-root/cbl directory and let me know if any of them qualify as powertools that should be moved to the new site-root/powertools/cbl directory.


site-root/cbl/powertools/backfill-missing-ratings.php
```diff
@@ -1,279 +0,0 @@
...
-            </label>
-            <label>
-                Max Date
-                <input type="date" name="maxDate" value="<?=$maxDateValue?>"/>
-            </label>
-            <button>Filter</button>
-        </form>
-    </body>
-
-</html>
```
themightychris: @BillClinton looks like this file became empty at the new location

themightychris (changes requested):

site-root/cbl/powertools/backfill-missing-ratings.php
BillClinton: odd.  sorry I missed that.

themightychris (approved):

PR #648, opened 2022-02-10
Show performance type if no task link [SCHOOL-155]

BillClinton:
This will show the value of the Demonstration.PerformanceType field in the Performance Task column of the skill list when the Demonstration does not have an associated StudentTask


BillClinton:
@themightychris It doesn't feel like this should be linkable as it is when the demonstration has an associated StudentTask, but let me know if you feel otherwise.

themightychris (approved):
looks good!

PR #652, opened 2022-02-15
feat(student-progress): hide unenrolled competencies [SCHOOL-147]

BillClinton:
- Closes SCHOOL-147

If a student is enrolled in some but not all competencies within a competency area,  hide the competency boxes that have no level available.

### DIFF
--- a/sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
+++ b/sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
@@ -150,9 +150,6 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
             studentCompetencyIndex, studentCompetency, level, competencyId,
             competency, competencyCurrent, average, growth,
 
-            // determine enrollment status by counting competencies
-            studentEnrolled = studentCompetenciesCount > 0,
-
             competencyLevels = {},
             lowestLevel = Infinity,
             totalRequired = 0,
@@ -239,6 +236,9 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
             }
         }
 
+        // set enrollment status
+        me.getDashboardCt().setHasEnrollments(cardConfigs.length>0);
+
 
         // update interface
         Ext.suspendLayouts();
@@ -256,9 +256,6 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
         cardsCt.add(cardConfigs);
 
         Ext.resumeLayouts(true);
-
-        // show/hide components based on enrollment
-        me.getDashboardCt().setHasEnrollments(studentEnrolled);
 
         // finish load
         competenciesSummary.setLoading(false);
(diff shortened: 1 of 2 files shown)


sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
```diff
@@ -222,17 +219,26 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
             }
         }
 
+        // get an array of competency IDs from the studentCompetenciesStore
+        let studentCompetencyIDs = studentCompetenciesStore.getData().items.map(rec => rec.get('CompetencyID'));
```
themightychris: **suggestion (non-blocking)**: I think you can just do this. Also note preferring `const` when a variable doesn't need to be reassigned

```suggestion
        const studentCompetencyIDs = studentCompetenciesStore.collect('CompetencyID');
```

sencha-workspace/SlateDemonstrationsStudent/app/controller/Dashboard.js
```diff
@@ -222,17 +219,26 @@ Ext.define('SlateDemonstrationsStudent.controller.Dashboard', {
...
         competenciesCount = competenciesStore.getCount();
         competencyIndex = 0;
 
         for (; competencyIndex < competenciesCount; competencyIndex++) {
-            cardConfigs.push({
-                competency: competenciesStore.getAt(competencyIndex)
-            });
+            let competency = competenciesStore.getAt(competencyIndex);
+
+            if (studentCompetencyIDs.includes(competency.get('ID'))) {
```
themightychris: **suggestion (non-blocking)**: Prefer the dedicated identifier getter for turning model instances into their identifiers (it respects the configured `idProperty`):

```suggestion
            if (studentCompetencyIDs.includes(competency.getId())) {
```

sencha-workspace/SlateDemonstrationsStudent/app/view/Dashboard.js
```diff
@@ -170,16 +170,18 @@ Ext.define('SlateDemonstrationsStudent.view.Dashboard', {
         me.fireEvent('selectedcontentareachange', me, contentArea, oldContentArea);
     },
 
-    updateHasEnrollments: function(enrolled) {
```
themightychris, line 173: **question (non-blocking)** Does this all need to re-run every time we call `setHasEnrollments`, even if the passed value hasn't changed?

The answer could be "yes" if any of the subcomponents this toggles the visibility of also have other ways they could get their visibility toggled from—that would necessitate an "enforce this state every time" approach that you get with `apply`. 

If this config is the authoritative boss of the visibility of all these subcomponents though, we can use the `update` method to get "only apply this state when the input changes". In a case where the config handler only relays values to a bunch of sub-component config handlers the difference doesn't matter much and you can err on the side of pushing the state out, but its always a good idea to minimize layout calls

If you had a reason using apply instead improved experience, then disregard this. It might be good practice to leave a comment why apply needs to be used since someone might think to "fix" it later

themightychris (approved):

themightychris:
Looks good in testing!

themightychris (approved):

PR #654, opened 2022-03-05
fix: export calculations consistently

nbey:
This should resolve an issue seen here: https://github.com/b21network/b21-skeleton/runs/5426475484?check_suite_focus=true#step:10:554
Our cypress test was previously interpreting empty strings as `0`, because the CSV export would not show `0` for Growth (the ternary check would cause calculations that = `0` to render as null.

Also, this should prevent an issue where NULL baselines are converted to `0` (`round()`ing a null value returns `0`)


themightychris (approved):

cypress/integration/student-competency-calculations.js
```diff
@@ -37,17 +37,17 @@ describe('Confirm rounding is consistent across UI, API, and exports', () => {
                             cy.wrap(xhr).its('status').should('eq', 200);
                             //convert api baseline to string, if it is not null
                             expect(
-                              apiBaseLine ? `${apiBaseLine}` : apiBaseLine,
```
themightychris, line 40: **nitpick**: We shouldn't transform what we pass into `expect`—we want to pass in the output we're checking as closely as possible to how we get it. Instead, transform what we pass to the assertion at the end of the statement

themightychris (changes requested):
shift all the transforms from the expects to the assertions

cypress/integration/student-competency-calculations.js
themightychris: Merging for now, but SCHOOL-164 captures following up on this.

PR #656, opened 2022-03-10
Show archived tasks when deep linking [SCHOOL-148]

BillClinton:


themightychris (approved):
@BillClinton perfectly done, and I really appreciate the great commit messages explaining the meaning of each change! Those are very helpful when reviewing

My only tweak is in the future, use the `style:` prefix for a commit that just changes namespace, `refactor:` is for commits that focus on rearranging existing code

PR #660, opened 2022-03-22
feat: hide deactivated students in tasks dashboard [SCHOOL-153]

BillClinton:
Students flagged as disabled should not appear in the tasks dash.  Also, create a new test suite that includes some disabled users in the fixture data, and use that to verify this fix.

TODO: 
- write tests


BillClinton:
This is not complete, but I am going to request a review in order to get feedback.  Technically it functions, but I'm not comfortable with the way I implemented it.  Specifically, I found that when I filtered the student-tasks endpoint to remove students that had been disabled, the column for that student would still appear in the UI with no data because the list of students is pulled from the section-participants endpoint.  The SectionParticipantsRequestHandler.php endpoint is a part of Slate, not slate-cbl.  As a proof of concept for the fix, I created a Slate/Courses/SectionParticipantsRequestHandler.php override in slate-cbl which uses the same filtering I implemented in the StudentTasksRequestHandler.   again, it works, but I am unsure how to approach a final implementation for the fix.

php-classes/Slate/CBL/Tasks/StudentTasksRequestHandler.php
```diff
@@ -501,6 +528,19 @@ public static function getRequestedArchiveFilter($fieldName = 'include_archived'
         return !!$_REQUEST[$fieldName];
     }
 
+    public static function getRequestedStudentDisabledFilter($fieldName = 'include_deactivated_students')
```
themightychris:
```suggestion
    public static function getRequestedIncludeDeactivated($fieldName = 'include_deactivated')
```

let's rename this param to `include_deactivated` so we can be consistent with the name of this filter across endpoints which are not all scoped to students. Deactivation applies to profiles of all types of which student is only one

Also, move this method up to `\Slate\RecordsRequestHandler`

php-classes/Slate/Courses/SectionParticipantsRequestHandler.php
```diff
@@ -0,0 +1,96 @@
...
+              } else {
+                  $conditions['PersonID'] = $excludeDisabled;
+              }
+          }
+      }
+
+        return $conditions;
+    }
+
+    public static function getRequestedStudentDisabledFilter($fieldName = 'include_deactivated_students')
```
themightychris: Use the method moved to `\Slate\RecordsRequestHandler`

php-classes/Slate/Courses/SectionParticipantsRequestHandler.php
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        // apply disabled students filter
+        if (!static::getRequestedStudentDisabledFilter()) {
+          $disabledStudentIds = \DB::allValues(
+              'ID',
+              '
+                  SELECT ID
+                    FROM `%s`
+                  WHERE AccountLevel = "disabled"
```
themightychris: `"Disabled"` is capitalized in the enum. I'm torn on capitalizing enum values and have stopped doing that but let's keep consistent with the existing enum values so we're never caught by case sensitivity issues

php-classes/Slate/Courses/SectionParticipantsRequestHandler.php
```diff
@@ -0,0 +1,96 @@
...
+
+          if (!empty($disabledStudentIds)) {
+              $excludeDisabled = [
+                  'operator' => 'NOT IN',
+                  'values' => $disabledStudentIds
+              ];
+              $existingPersonIDCondition = $conditions['PersonID'];
+
+              if (isset($existingStudentIDCondition)) {
+                  $conditions['PersonID'] = [$existingPersonIDCondition, $excludeDisabled];
```
themightychris: Can you confirm this works? I don't recall mapConditions supporting multiple conditions.

In the case of a condition you want to append but aren't intentionally trying to make it replaceable by downstream filters on the same field, it's best to just not key it by a field at all, like you can see here: https://github.com/JarvusInnovations/emergence-skeleton/blob/develop/php-classes/PeopleRequestHandler.class.php#L77

php-classes/Slate/CBL/Tasks/StudentTasksRequestHandler.php
```diff
@@ -60,6 +61,32 @@ protected static function buildBrowseConditions(array $conditions = [], array &$
             ];
         }
 
+        // apply disabled students filter
+        if (!static::getRequestedStudentDisabledFilter()) {
+            $disabledStudentIds = DB::allValues(
```
themightychris: I don't looovee this approach of getting a list of IDs first to exclude, but after thinking about it a bit I do think it's our best option, and this is a pattern we use in a lot of places.

What I'd like to do though is create a public static method on [Slate's override of the `Person` model class](https://github.com/SlateFoundation/slate/blob/develop/php-classes/Emergence/People/Person.php) like `Person::getDeactivatedIds()`

Then that method can cache the ID list like [this method does](https://github.com/SlateFoundation/slate-cbl/blob/develop/php-classes/Slate/CBL/Competency.php#L121)—except in this case we don't need to include any ID in the cache key, it could be a static global key like `deactivated-people-ids`. And then you can have an after-save handler like this that just blows out that whole key any time an `AccountLevel` field changes [like this method does](https://github.com/JarvusInnovations/emergence-skeleton/blob/develop/php-classes/Emergence/CMS/ContentBlock.php#L76)

The list of deactivated account IDs won't change frequently and we'll want to use it in a lot of filters so I think it makes sense to let the `Person` model handle caching a static ID list for us

themightychris (changes requested):
Good approach overall, suggested some tweaks to the factoring. Looks like we'll need to start with a PR to the `slate` repo, and lets in that PR try to take stock of any places in slate that are calling that section participants API endpoint to confirm we want deactivated accounts being excluded and include a test for it at that level

php-classes/Slate/Courses/SectionParticipantsRequestHandler.php
BillClinton: It worked when I tested it, but I like the idea of adding the condition the way your example does it better anyway.  There's much less code if I don't have to worry about a StudentID condition already being set.

php-classes/Slate/Courses/SectionParticipantsRequestHandler.php
```diff
@@ -0,0 +1,67 @@
+<?php
```
themightychris: @BillClinton this override needs to be deleted now that slate#339 is merged and this branch has been rebased on top of upgrading slate to the release that includes it

themightychris (changes requested):

php-classes/Slate/Courses/SectionParticipantsRequestHandler.php
themightychris: @BillClinton i got it

themightychris:
@BillClinton see [1e8a530](https://github.com/SlateFoundation/slate-cbl/pull/660/commits/1e8a5309d01bdd0184f0a9fc6391b7e9cbcc90a2) — the problem with the initial implementation here was it triggered both stores to reload and creates a janky flicker as the grid reloads twice, you can see it in the console:

Screen Shot 2022-05-24 at 3 30 57 PM

The change I made to call `.unload()` on both first (a method we add to every store) resets it to the unloaded state, and the aggregrid component knows not to re-render until both stores are in a loaded state

PR #675, opened 2022-04-23
feat(growth): provide original growth behavior

themightychris:


themightychris:
Closing this effort as all schools using the original PLMB calculation have accepted the updated version

PR #671, opened 2022-04-11
Tests/adding wait/assert for xhr [SCHOOL-144]

Camus1859:

PR #678, opened 2022-04-25
Tests/adding wait/assert for xhr calls in tasks manager

Camus1859:


cypress/integration/cbl/tasks_manager.js
```diff
@@ -7,8 +7,8 @@ describe('CBL: Tasks Manager Test', () => {
 
     // authenticate as 'teacher' user
     beforeEach(() => {
-        cy.server().route('GET', '/cbl/tasks').as('taskData');
-        cy.server().route('POST', '/cbl/tasks/save*').as('taskSave');
+        cy.intercept('GET', '/cbl/tasks?!(\\?*)').as('taskData');
+        cy.intercept('POST', '/cbl/tasks/save?!(\\?*)').as('taskSave');
```
themightychris: @Camus1859 `?!(\\?*)` doesn't seem to be doing anything—`?!` is not a documented prefix with any meaning I can find here: https://github.com/motemen/minimatch-cheat-sheet

So I think this pattern is saying:
- There isn't `?*` at the end
- The above may or may not be true

Looking at actual calls being made during this test, these requests do have misc `?query=params` appended that we want to allow for but not make a hard requirement, so the pattern I showed in #671 is more appropriate

themightychris (approved):
@Camus1859 good call keeping the destroy intercept only in one test since we only use it there, and making the other two shareable.

I'm going to go ahead and push the pattern fix I suggested since I have Cypress running on this branch already and get it merged, please take note and review the minimatch cheatsheet for the future

PR #683, opened 2022-05-22
chore: added editorconfig and prettier to improve formatting for Cypress test[SCHOOL-119]

Camus1859:
From the root, enter the following command to display and fix eslint errors and to format with prettier 
`npx eslint --fix cypress/**/*.js && prettier --write cypress/**/*.js`

I was thinking of making the line above an `npm` script but there is no `package.json` but we obviously have one somewhere since we are using dependencies like Cypress... Maybe someone can help clarify how this is happening..


package-lock.json
```diff
@@ -0,0 +1,31 @@
+{
```
themightychris, line 1: we need to delete this from this PR, any npm deps need to be installed upstream in the emergence-skeleton repo where `package.json` and the rest of the base cypress environment live

in fact... everything in this PR should be going there instead, please close this PR and reopen one there. Changes up `package.json` and `package-lock.json` should all be applied through running `npm` commands in that repo.

After that PR merges I'll handle running the reformat up the whole stack as I bump versions

themightychris (changes requested):

PR #684, opened 2022-06-07
Improve task library grid [SCHOOL-162]

BillClinton:
Make improvements to the Task Manager Grid including: 
 * Configure Sencha’s built-in [grid filter plugin](https://docs.sencha.com/extjs/6.2.0/classic/Ext.grid.filters.Filters.html)

 * Fix joining of skills into comma-delimited list

 * Implement reasonable sorter functions for most/all columns

 * Make sure paging works

 * Implement a resizable split view between grid and right-side details panel


BillClinton (commented):
I made a note about the two race conditions that are causing test failures.  We can be pretty sure that these are race conditions because if I duplicate the previous focus command so it is performed twice, there are no test failures. @themightychris

php-classes/Slate/CBL/Tasks/Task.php
```diff
@@ -233,4 +262,135 @@ protected static function validateTaskStatus(RecordValidator $RecordValidator, T
...
+    public static function getTitleConditions($identifier, $matchedCondition)
+    {
+        return static::getTableAlias().'.Title LIKE "%%'.$identifier.'%%"';
+    }
+
+    public static function getParentTaskConditions($identifier, $matchedCondition)
+    {
+        $parentTasks = DB::allRecords('SELECT ID FROM %s WHERE Title LIKE "%%%s%%"', [
+            Task::$tableName,
+            $identifier
```
themightychris, line 275:
```suggestion
            DB::escape($identifier)
```

php-classes/Slate/CBL/Tasks/Task.php
```diff
@@ -233,4 +262,135 @@ protected static function validateTaskStatus(RecordValidator $RecordValidator, T
...
+        $relatedTasks = DB::allRecords('
+            SELECT ts.TaskID
+            FROM %s as ts
+            JOIN %s as sk
+              ON ts.SkillID = sk.ID
+            WHERE sk.Code LIKE "%%%s%%"',
+        [
+            TaskSkill::$tableName,
+            Skill::$tableName,
+            $identifier
```
themightychris, line 298:
```suggestion
            DB::escape($identifier)
```

php-classes/Slate/CBL/Tasks/Task.php
```diff
@@ -233,4 +262,135 @@ protected static function validateTaskStatus(RecordValidator $RecordValidator, T
...
+            FROM %s as task
+            JOIN %s as person
+              ON task.CreatorID = person.ID
+            WHERE person.FirstName LIKE "%%%s%%"
+            OR person.LastName LIKE "%%%s%%"',
+        [
+            Task::$tableName,
+            Person::$tableName,
+            $identifier,
+            $identifier
```
themightychris, line 327:
```suggestion
            DB::escape($identifier),
            DB::escape($identifier)
```

themightychris (changes requested):
@BillClinton looks great, fabulous work!

The only major issue I see is that we're putting user input directly into filter SQL without escaping, that will need fixing before merge

sencha-workspace/SlateTasksManager/sass/src/view/TasksManager.scss
```diff
@@ -1,9 +1,23 @@
...
+}
+
+#slateapp-viewport {
+  height: 100%;
+  position: absolute;
+  top: 0px;
+  bottom: 0px;
+  width: 100vw;
+}
+
```
themightychris: in stuff within `app/sass/{src,var}`, each file corresponds to a component and it's best to avoid including any CSS anywhere in this tree that wouldn't be scoped to the component the file is for. Notice the version before this only had stuff within `.slate-tasks-manager`

`app/sass/etc/all.scss` is the place for random global stuff

In any case, I found this calculation to be off a bit and leaving gaps in the window, so I've replicated how SlateAdmin over in the `slate` repository is set up and did some other clean up: https://github.com/SlateFoundation/slate-cbl/pull/684/files/b6b42921ddcf39a45ec8d20e35fa007c7091f860..e7544db3d673dbfa063bac38ad6d8d3c65242a21

themightychris (approved):

PR #698, opened 2022-08-16
feat: add dev documentation for task manager app [SCHOOL-188]

BillClinton:


BillClinton:
@themightychris  I used the SlateAdmin dev docs as a template, but I am not sure if the part about `git submodule update --init` is necessary or if a `fetch && pull` should be assumed

themightychris (approved):

PR #708, opened 2022-09-21
Issue 706/vue state management

chriscauley:
## Usage

* If working locally, start studio and then `cd static-webapps/slate-portfolio-manager` and `npm start`.
* If running not local, set I believe setting `BASE_URL=/` will work to hook up the redirects and api calls. The rest of this tutorial will reference the local build
* Go to localhost:8080 and you should be redirected to the login page on the api sever (localhost:2190). Login with admin:admin and you should be redirected back to localhost:8080
* Select "English Language Start" and "Class of 2021" and the table should update.
* Click the top left cell of the table (should be "clarissa cross" and "reading critically")
* Collapse the Year 12 and toggle "show hidden items" (the toggle will not appear if student has no hidden items)

## Centralized http client

In store/client.js exports an instance of axios that redirects to the login page if any request returns a 401 response and standardizes requests (Accept header and credentials). For the most part the app won't use this directly.

## Restful state management

All of the pinia modules currently use `store/defineRestStore` to create reactive stores that can be used in components. The most basic usage is as follows

``` javascript
<template>
    <div v-if="student">{{ student.Username }}</div> <!-- note 4 -->
</template>

<script>
import { mapStores } from 'pinia'
import useStudent form '@/store/useStudent' // note 1

export default {
    computed {
        ...mapStores(useStudent), // note 2
        student() {
            const { studentName } = this.$router.query // eg ?studentName=bob
            const response = this.studentStore.get(`?student_name=${studentName}`) // note 3
            return response && response.data // note 4
        }
    }
}
</script>
```

1. Stores require an id and a url. Very basic usage can be seen with useStudent, more advanced usage with useAuth
2. This gives the component access to `this.studentStore`
3. This will kick off a request, return undefined for now, and then the page will re-render once the request completes. If you need/prefer to use async/await you can do `this.studentStore.fetch(target)`.
4. Because this uses reactivity instead of async/await, the get call returns undefined if the cache is empty. This means templates need to be guarded with `v-if="student"` so that rendering `{{ student.Username }}` doesn't throw an error

If two components make the same call of `this.StudentStore.get(target)`, only one request will be called. ~~This PR does not include cache invalidation, so requesting the same resource (eg clicking back and forth between two students in the) will retrun stale data.~~

### Refetching and marking data as stale

I added an update function to the `defineRestStore` api. All this does is post the first argument to `baseUrl + '/save'`. Afterwards you can call `this.whateverStore.get(whateverUrl, true)` where true tells the store to fetch even though it has cached data for that url.

## Auth State management

Typically I prefer to do this in a `router.beforeEach` request, but I couldn't get pinia to work outside of components with Vue 2. We want the auth state to fully load before we load the page (and start kicking off all sorts of auth-protected api calls). So src/views/AdvancedPortfolioManager.vue uses `async mounted() { await this.authStore.required() }` and `v-if="authStore.user"` to guard the rest of the view from loading.

## Local storage state (ui)

I called this "ui" but it will probably be broken into multiple submodules as this app grows. Any state stored in this will persist in local storage (eg refreshing the page will keep the side bar open to a given student).

# Models

I like to add business logic to a models folder. Right now it only contains a method for printing the student name, but I find making a folder like this very useful to aggregate logic.
 
## Colors!

I'm pretty proud of my solution to the color problem. If you add `cbl-level-10` class to any element that sets a series of color variables. In that element or any descendant you can then use `bg-cbl-level` to make the background the color or `bg-cbl-level-50` to make it the 50/50 blend of the color and white (50% opacity in the design file).

## Next steps

* Usually I'd make more high level requests on a store. Rather than calling `studentStore.get('?student="+studentName)` I'd make methods like `studentStore.getStudent(studentName)` and `studentStore.getMany({ contentArea, studentList })`. This way any shared functionality (ie adding things like "limit=0") can exist in the storage rather than the component.
* ~~Currently there's no way to reject stale data. If anyone knows of a easy post request I could add let me know and I'll add it. typically I'd add a `studentStore.save` method that will make the request, mark the data stale, and update it.~~ Most recent commit handles saving, refetching, and detecting a loading state.


themightychris:
@chriscauley I've rebased this PR to incorporate #709 and adjusted `baseUrl` such that `/` is the default so that integrated builds use their host environment by default, and then pointing it at an external place like `localhost:2190` is the exception you need to override for in development environments.

Upon trying to implement this I found that Vue will only pass `VUE_APP_*` environment variables through to client-side instances of `process.env.*` except for `BASE_URL` which has a special meaning in Vue and is driven by `vue.config.js`

So I added `VUE_APP_API_BASE=http://localhost:2190/` to `.env.development` and checked that into source control, which should achieve overriding that value during local dev without interfering with production builds.

themightychris:
Some initial takeaways from demo:

Fix before merge:

- [x] UI grayout needs to span both the save and reload requests to prevent users from rapidly queuing overlapping changes
- [x] Arrows need to be correctly colored, also put an HTML label on them that states "Move to level X"
- [x] Availability of arrows should be restricted based on what levels are already on the screen (we will force users to do two discreet steps if they want to create a new level and then move stuff there)
- [x] Add a confirmation prompt between click and write that has OK focused

Follow-up tasks after merge: 

- [ ] We need to create a test that simulates a level shift that triggers graduation, and ensure that everything (including the main grid) update correctly—#713
- [ ] See if Ryon can layer in some animation to make it clearer what change happened—#712

chriscauley:
Here's a list of some "edge cases" that don't look like the other rows. The first row is my description, second row is how to find it (competency area dropdown + students dropdown + column + row), and the third row is what effect it has on the table.
* Demonstrations with StudentTasks
   - ELA + class of 2020 + Student Slate + reading critically
   - shows task Title instead of demo Context
* "Missing" (DemonstratedLevel === 0)
   - ELA+ 2020 + Stuent2 Slate + Expressing Ideas shows gray "M" levels
   - shows "M" and is grayed out
* Hidden items
   - ELA + 2021 + Sherill + Reading Critically
   - Toggle "show hidden items" to see some grayed out items
* Overridden
  -  ELA + 2021 + Sherill + Reading Critically
  -  Several items have checkmarks and "[Overridden]" instead of a demo description

PR #722, opened 2022-10-29
adding ability to delete portfolio

chriscauley:
This probably needs some copy tweaks. Right now the backend cache isn't being reset after you delete a portfolio. The first deletion seems to work fine (the server returns "success: true") but when the front end tries to refetch the data, the portfolio still exists. If you refresh the page, the portfolio is still there, but when you try to delete it you get a 404 error (so the first attempt must have been successful).

Below are some images showing what the delete button, "are you sure" prompt, and 404 error look like.

image

image

### DIFF
--- a/php-classes/Slate/CBL/StudentCompetenciesRequestHandler.php
+++ b/php-classes/Slate/CBL/StudentCompetenciesRequestHandler.php
@@ -105,4 +105,14 @@ class StudentCompetenciesRequestHandler extends RecordsRequestHandler
 
         return $conditions;
     }
+
+
+    protected static function onBeforeRecordDestroyed(ActiveRecord $StudentCompetency)
+    {
+        if (count($StudentCompetency->getDemonstrationsData()) > 0) {
+            return static::throwInvalidRequestError('Portfolios must be emptied before they are deleted');
+        }
+
+        return parent::onBeforeRecordDestroyed($StudentCompetency);
+    }
 }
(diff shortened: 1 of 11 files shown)


themightychris:
@chriscauley in 89c6f795fa37351942877723c3e0fbb60da6573f I added a server-side error for nonempty portfolios: 
Screen Shot 2022-11-01 at 12 59 44 PM

Looks like this is handled well enough by the UI already, I tested by logging a demonstration to the selected portfolio in a separate tab after I already opened it and got the delete button in front of me

Screen Shot 2022-11-01 at 2 58 38 PM

themightychris:
@chriscauley after deleting a topmost portfolio, the side pane refreshes but the cell in the main grid doesn't update until you hard reload the page:

## After delete

Screen Shot 2022-11-01 at 2 52 30 PM

## After hard refresh

Screen Shot 2022-11-01 at 2 52 41 PM

themightychris (changes requested):
Looks like the only issue is the grid refresh I described here: https://github.com/SlateFoundation/slate-cbl/pull/722#issuecomment-1298967446

chriscauley:
> @chriscauley after deleting a topmost portfolio, the side pane refreshes but the cell in the main grid doesn't update until you hard reload the page:

This is complicated. I'm going to outline my solution here before I start implementing it. Feel free to give me feedback.

The issue is that those are pulling from different parts of the API so the side bar needs to understand that the enrollments grid is on the page and have the ability to trigger a refetch.

I think the best solution is going to be to use an event emitter. So when the sidebar changes a student it will emit an event like `emitter.emit('update-student', student.ID)`. The enrollment grid will listen, see the event, recognize student.ID as being a student in the table, and refetch the data.

This way as the site grows the individual components will just need to know what data they are currently displaying and to listen to the emitter for events regarding that data.

Issue #731, opened 2022-11-08
No portfolio visible if only portfolio is empty


themightychris:
## Steps to reproduce

1. Open ELA+Example school with fixture data loaded: <http://localhost:2190/webapps/SlatePortfolioManager/#/?area=ELA&students=group%3Aexample_school>
2. Select any of the level-9 cells under Edmund Ebel's column after the first row

## What happened

Nothing renders despite a single portfolio (with no demonstrations) coming back in the load response:

![Screen Shot 2022-11-08 at 1 15 48 PM](https://user-images.githubusercontent.com/458494/200643569-c2e43396-0989-49e7-9c23-6c2a51f3e7f6.png)

## What should happen instead

I should see an empty level 9 portfolio with a "Delete Portfolio" button available. Being able to delete the only portfolio is useful in cases where a student was mistakenly enrolled in too-high an initial portfolio. Clearing out that single empty portfolio enables the user to then use the old enrollments dashboard to re-enroll them at a lower starting level

themightychris:
Fixed in #733

PR #733, opened 2022-11-11
fix(portfolio-manager): guard missing demonstration at the SkillDemos level [#731]

chriscauley:


themightychris (approved):

themightychris:
Verified this attains the desired behavior:

PR #746, opened 2022-11-28
feat(portfolio-manager): add enroll buttons [#736]

### LINKED ISSUE #736: Implement button to create enrollment in next portfolio
Currently in [original enrollments manager](https://latest.slate-cbl.sandbox.k8s.jarv.us/cbl/dashboards/student-competencies/admin#ELA/group:example_school) you can create an enrollment anywhere between a site's configured min and max levels (typically 9-12 but this should be a top-level config we can drive from the backend later). In cells with no existing enrollment you can create one starting at the min level (e.g. 9) and in cells with an existing highest enrolled portfolio level you can create an enrollment in the next or any higher level. If a student currently has a level 9 portfolio for example, and you click through to create a level 12 enrollment, only a level 12 enrollment gets posted to the server and that student then has a sparse stack of portfolios that jumps from whatever level they were at right up to 12 (i.e. we don't need to and perhaps want to avoid creating empty intermediary portfolios if the admin's goal is to jump them forward)—this behavior varies a bit from what we discussed on the call in that we mentioned only needing to let people jump to the immediate next level and to let them jump one at a time if they need to go higher

## Proposed initial implementation

- For every available portfolio level between the currently highest one and the max (i.e 12), show a collapsed and ghosted out portfolio header at the top of the side panel with an "Enroll" button in place of the progress bar
- When selecting an empty cell with NO current portfolio enrollments, given a global config of min=9 and max=12, you should see 4 such headers for levels 9-12 with enroll buttons
- We do not want to "fill" the sidebar in with enroll buttons for intermediary portfolios—they should only be made available above the highest already-enrolled portfolio
    - example: if the student has a level 9 portfolio and a level 12 portfolio there will be no enroll buttons
    - example: if the student has a level 9 portfolio and a level 11 portfolio there would be one enroll button available for level 12 at the top of the side panel
- There is no need to show a modal confirmation prompt for this action since it can easily be immediately undone with no side-effects via the "Delete portfolio" button

chriscauley:
Fixes #736 

* Created `this.$site`, a global variable for max and min level (currently hardcoded to 9 and 12)
* Show every level above students current level
* If student is in `$site.maxLevel` show nothing.
* If student is not in any portfolio, show levels from `$site.minLevel` to $site.maxLevel`

Demonstration

[enroll-button.webm](https://user-images.githubusercontent.com/379151/204341511-76b325b6-b667-406b-8237-a59f4760b2b2.webm)

### DIFF
--- a/static-webapps/slate-portfolio-manager/src/components/sidebar/NewLevelPanel.vue
+++ b/static-webapps/slate-portfolio-manager/src/components/sidebar/NewLevelPanel.vue
@@ -1,0 +1,64 @@
+<template>
+  <div
+    :class="`level-panel -new mb-2 cbl-level-${Level}`"
+  >
+    <b-container class="bg-cbl-level-25">
+      <b-row class="py-2 align-items-center">
+        <b-col>
+          <h3 class="h6 m-0 text-muted">
+            Year {{ Level }}
+          </h3>
+        </b-col>
+        <b-col cols="4">
+          <button
+            class="btn btn-primary btn-sm float-right"
+            @click="createLevel"
+          >
+            Enroll
+          </button>
+        </b-col>
+      </b-row>
+    </b-container>
+  </div>
+</template>
+
+<script>
+import client from '@/store/client';
+import emitter from '@/store/emitter';
+
+export default {
+  props: {
+    Level: {
+      type: Number,
+      default: () => null,
+    },
+    StudentID: {
+      type: Number,
+      default: () => null,
+    },
+    CompetencyID: {
+      type: Number,
+      default: () => null,
+    },
+  },
+  methods: {
+    createLevel() {
+      const { StudentID, CompetencyID, Level } = this;
+      const data = [
+        {
+          StudentID,
+          CompetencyID,
+          Level,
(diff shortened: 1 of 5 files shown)


themightychris (approved):

static-webapps/slate-portfolio-manager/src/components/sidebar/NewLevelPanel.vue
```diff
@@ -0,0 +1,64 @@
...
+          StudentID,
+          CompetencyID,
+          Level,
+          EnteredVia: 'enrollment',
+          ID: -1,
+          Class: 'Slate\\CBL\\StudentCompetency',
+          BaselineRating: null,
+        },
+      ];
+      const url = 'cbl/student-competencies/save';
```
themightychris, line 58: @chriscauley this fails under hosted builds, I didn't think to test it out in the PR environment build. I think we need a leading `/` here—I'm gonna try that out and just go ahead and commit and finish the release if it works, just tagging you for future reference

Screen Shot 2022-11-29 at 6 54 28 PM

PR #747, opened 2022-11-28
feat(portfolio-manager): add unstyled drag-and-drop demo [#740]

### LINKED ISSUE #740: Implement drag-and-drop movement of demonstration-skills between portfolios/levels
Currently, you can use the up/down arrow buttons to move a demonstration-skill record between portfolios.

Screen Shot 2022-11-15 at 3 03 56 PM

It is intended that you should also be able to grab one of the rows via the grey bar (to be restyled in #739) to drop into another portfolio, potentially skipping by multiple. Once you pick up a row, the available drop targets should be all other portfolios on the screen. The target can only be a portfolio as a whole, there is no value in being able to see or target any particular elements or positions within each portfolio. Therefor it has been suggested that when the drag starts, a CSS class be applied to the overall sidebar indicating that the panel is in drop mode, which will switch the UI into a state where every portfolio is reduced to a simple mid-sized drop target such that you can see all of them on the screen at once, and the only exposed detail for each needs to be the color and year/level number. The existing portfolio or outside any other hit target would cancel the drop, and any other portfolios would be valid drop targets.

Ryon can refine the styling later, an initial pass that puts CSS classes in the right places and comes with enough styling to make the interaction obvious is sufficient to close this ticket.

chriscauley:
Fixes #740. Skill demos can now moved between levels using the mouse.

* Dragging a skill demo collapses all the children of the skill demo (bottom lighter section is empty)
* Dropping it moves between levels
* Dropping on original level has no effect
* State management functionality added to src/store.useUi.js

Demonstration:

[drag-skill-demo.webm](https://user-images.githubusercontent.com/379151/204358189-a6a40f28-cf30-4177-b5bd-02405abddf2d.webm)

### DIFF
--- a/static-webapps/slate-portfolio-manager/src/assets/app.scss
+++ b/static-webapps/slate-portfolio-manager/src/assets/app.scss
@@ -37,3 +37,15 @@ body {
     z-index: 99999;
   }
 }
+
+.level-panel.-drop-zone {
+  position: relative;
+  &:before {
+    background: green;
+    content: "";
+    display: block;
+    inset: 0;
+    position: absolute;
+    z-index: 1;
+  }
+}
--- a/static-webapps/slate-portfolio-manager/src/assets/app.scss
+++ b/static-webapps/slate-portfolio-manager/src/assets/app.scss
@@ -39,13 +39,7 @@ body {
 }
 
 .level-panel.-drop-zone {
-  position: relative;
-  &:before {
-    background: green;
-    content: "";
-    display: block;
-    inset: 0;
-    position: absolute;
-    z-index: 1;
+  .btn-danger, .skill-demos {
+    display: none;
   }
 }
--- a/static-webapps/slate-portfolio-manager/src/assets/app.scss
+++ b/static-webapps/slate-portfolio-manager/src/assets/app.scss
@@ -39,7 +39,11 @@ body {
 }
 
 .level-panel.-drop-zone {
-  .btn-danger, .skill-demos {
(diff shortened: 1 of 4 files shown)


tgaffey:
I was testing dragging and dropping and in the middle of grabbing a rating, I actually forgot where i grabbed it from. is there anything simple we can do to make this a little more obvious? Im thinking of two things. Can the level be highlighted when hovering over it with a rating? Can the icon change if you are hovering over the level it came from? Something visual to make it clear if you try to drop a rating into the initial level.

themightychris:
> I was testing dragging and dropping and in the middle of grabbing a rating, I actually forgot where i grabbed it from. is there anything simple we can do to make this a little more obvious? Im thinking of two things. Can the level be highlighted when hovering over it with a rating? Can the icon change if you are hovering over the level it came from? Something visual to make it clear if you try to drop a rating into the initial level.

Let's log this in our follow-up ticket for Ryon to apply refined styling

@chriscauley for the scope of this PR can you just make sure the source level container and all the potential destination level containers have distinct CSS classes applied designating something like `.drag-source` and `.drag-target`

themightychris (changes requested):
add applying drop source or drop target CSS class to every level container in the collapsed view so Ryon can implement visual effects to intuitively differentiate them in a later PR

chriscauley:
I added `-drag-source` and `-drag-target` classes and made the source element grayed and and 50% opacity. It doesn't look great, but it'll do until we get some design on it.

image

themightychris (approved):

PR #766, opened 2022-12-31
fix(student-progress-dashboard): hide unrendered elements before competency area selection [#765]

### LINKED ISSUE #765: Unrendered elements visible before selecting a competency area
## Steps to reproduce

- Login as `student`
- Open the [Competency Dashboard](https://latest.slate-cbl.sandbox.k8s.jarv.us/cbl/dashboards/demonstrations/student)
- View in default initial state without selecting a competency area

Screen Shot 2022-12-30 at 6 27 21 PM

## What I expect to see

Only the "Select a content area to load demonstrations dashboard" placeholder

## What I see instead

The placeholder, plus unloaded competency area summary, recent progress, and portfolios components

If you select a competency and the clear the selection, the page then looks like it should:

Screen Shot 2022-12-30 at 6 36 02 PM

## Thoughts

I feel like this must be a decently recent regression...maybe something happened with the changes in #652 ? Nothing obvious jumps out at me but it might be worth starting with a test of whether the bug still happens before and after that PR was merge

themightychris:
- Closes #765

### DIFF
--- a/cypress/integration/cbl/progress/student-dashboard.js
+++ b/cypress/integration/cbl/progress/student-dashboard.js
@@ -21,7 +21,9 @@ describe('CBL / Progress / Student Dashboard', () => {
 
         // verify placeholder is visible and none of its siblings
         cy.get('.slate-appcontainer-bodyWrap .slate-placeholder:visible')
-            .should('have.text', 'Select a competency area to load demonstrations dashboard');
+            .should('have.text', 'Select a competency area to load demonstrations dashboard')
+            .siblings(':visible')
+                .should('have.length', 0);
 
         cy.get('.slate-demonstrations-student-cardsct')
             .should('be.empty');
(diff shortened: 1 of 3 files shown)


themightychris:
@BillClinton this approach is kind of drifting outside of the preferred component lifecycle flow, and keeping that well-defined is really important to the long term stability of complex Sencha UIs

The `update*()` is where state transitions should be getting propagated. All the `me.set{ComponentName}(isVisible)` calls we are making are going through mechanisms specifically designed to work both before and after the rendering or initialization of those components.

It sounds like the issue was that `updateSelectedContentArea()` wasn't getting called initially? Did you try hard-coding a breakpoint into there to measure conclusively when and under what conditions it ran? I don't recall exactly, but I think that with it defaulting to `null` in the `config: { }` block would cause the `update*()` method not to run initially. I think if you use `true` or `false` as a default here instead that will cause it to run initially during component initialization before render. OR we should be calling `setSelectedContentArea()` somewhere during the relevant controller's launch lifecycle. If there isn't already a code path in the controller running and handling the case of initially loading with nothing selected, and you find that the `null` default means `updateSelectedContentArea()` doesn't run during initialization, see if changing its default from `null` to `false` triggers it running once during initialization

PR #772, opened 2023-01-23
fix(omnibar): ensure omnibar overlaps app content [#425]

### LINKED ISSUE #425: Selectors cover omnibar menu
When the content area or student group is selected in the competency dashboard, the dropdown menus persist and cover the omnibar menu when expanded. 

screen shot 2018-02-06 at 12 18 58 pm

mijogu:
Lowers the z-index for dropdown menus in the header so that the omnibar is on top

### DIFF
--- a/sencha-workspace/packages/slate-cbl/sass/etc/all.scss
+++ b/sencha-workspace/packages/slate-cbl/sass/etc/all.scss
@@ -7,3 +7,8 @@
     $level-colors-medium,
     $level-colors-dark
 );
+
+div[id^='slate-cbl-'][id*='-picker'],
+.x-css-shadow {
+    z-index: 10 !important;
+}


sencha-workspace/packages/slate-cbl/sass/etc/all.scss
```diff
@@ -7,3 +7,8 @@
     $level-colors-medium,
     $level-colors-dark
 );
+
+div[id^='slate-cbl-'][id*='-picker'],
+.x-css-shadow {
+    z-index: 10 !important;
+}
```
themightychris, line 14: It's not a good idea to monkey-patch the generated ExtJS styling like this, and the selector is liable to not catch everything ExtJS might put above the omnibar

A better approach instead of this would be to find the existing place where `.slate-omnibar` is assigned a `z-index: 99` and set that to `z-index: 999999` as our intention is that the omnibar should cover EVERYTHING. That way we don't have to get into whack-a-moleing specific things to be lower than it

themightychris (changes requested):

PR #777, opened 2023-02-07
fix(e2e): prevent race conditions during overrides tests

BillClinton:


cypress/integration/cbl/progress/overrides.js
```diff
@@ -177,7 +178,7 @@ describe('CBL / Progress / Overrides', () => {
                     .should('contain.text', '0%')
     });
 
-    it('Submit a rating for a skill with an override.', () => {
+    it.only('Submit a rating for a skill with an override.', () => {
```
themightychris: @BillClinton be careful about leaving `.only` in commits (I do it all the time but check your Files Changed tab after opening a PR)

PR #781, opened 2023-02-10
refactor: avoid extGet in within block

BillClinton:


cypress/integration/cbl/admin/tasks-manager.js
```diff
@@ -323,30 +323,31 @@ describe('CBL / Admin / Tasks Manager', () => {
...
-
-                    });
+                cy.get('input[name="ClonedTaskID"]')
+                    .should('have.length', 1)
+                    .parents('.x-form-trigger-wrap')
+                        .should('have.length', 1)
+                        .within(() => {
+                            cy.get('.x-form-arrow-trigger')
+                                .click()
+                        });
```
themightychris: There's no reason to use `.within()` if you're not using the scope more than once. This could just be:

```suggestion
                        .find('.x-form-arrow-trigger')
                            .click();
```

cypress/integration/cbl/admin/tasks-manager.js
```diff
@@ -323,30 +323,31 @@ describe('CBL / Admin / Tasks Manager', () => {
...
+                        const componentID = $input.attr('data-componentid');
+
+                        cy.document().its('body').find(`#${componentID}-picker-listEl`)
+                            .contains('.x-boundlist-item', 'A Shared Task')
+                            .should('exist');
+
+                        cy.document().its('body').find(`#${componentID}-picker-listEl`)
+                            .contains('.x-boundlist-item', 'An UnShared Task')
+                            .should('not.exist');
+                    })
```
themightychris: eh I think it was better before, the dialog is modal and the selectors were sufficiently unique. There's just no point in them being inside the `within` so maybe just revert and move them below the `.within` so it's just sequenced after we wait for the window to be visible and there's no big window-wide `.within()` since we don't ned to do any multi-element DOM-based scoping

themightychris (changes requested):

cypress/integration/cbl/admin/tasks-manager.js
BillClinton: I was questioning the necessity of doing it too.  I like the within aesthetically because it groups a bunch of assertions on elements in the form window, but then you have to bend over backwards with the dropdown because it's technically floating outside the window.  I put the original code back, but outside the within.

themightychris (approved):

cypress/integration/cbl/admin/tasks-manager.js
```diff
@@ -554,6 +539,12 @@ describe('CBL / Admin / Tasks Manager', () => {
...
                     })
             })
-        return items;
+            .then(() => {
+                if (reverse) {
+                    cy.expect(items, `Column [index: ${col}]: items are sorted in reverse alphabetical order.`).to.deep.equal(items.slice().sort().reverse());
+                } else {
+                    cy.expect(items, `Column [index: ${col}]: items are sorted in alphabetical order.`).to.deep.equal(items.slice().sort());
+                }
+            })
```
themightychris, line 548: I've never seen the `.then(() => { cy.expect(...) } ` pattern...seems like it _should_ work

Typically what I do in these cases is pass a function to `.should()` and use the bare `expect` function. I know that when you do that, the `.should()` wrapper catches the exceptions thrown by any failing assertions within and retries the whole `.should()` block up to the allowed timeout. 

The current implementation feels a bit odd with the shared state between the `.each()` and `.then()`. You could do all of this in a single `.should()` method which receives the subject element(s) already wrapped in jquery

I think you could just do something like:

```javascript
cy.get('.x-grid-row', { log: false })
            .should($rows => {
                expect($rows.length).to.be(20);
                const items = $rows.map($row => $row.find('.x-grid-cell-inner').text())
                expect(items)...;
            });

PR #786, opened 2023-04-06
fix: show previous portfolios for enrollment

BillClinton:
Currently the Portfolio Manage sidebar does not show any levels below the lowest level with a portfolio.  This means users cannot enroll in those levels.  This fix will change the behavior of the application, showing all levels, and enabling enrollment in those levels.

One note: a computed property "visibleLevels" exists in the AdvancedPortfolioSidebar component.  This property returns an array of the levels with portfolios.  Now that we are showing all levels in the sidebar, this property is no longer used in this component, but it is passed down to and used in subcomponents:

- LevelPanel: used to determine if a portfolio can be deleted - only topmost portfolios can be deleted.
- SkillDemoCard: used to determine whether levels are valid targets for drag-and-drop operations.

While the function of this property is the same, technically since we are now showing all levels, it might be better named levelsWithPortfolios or enrolledLevels something similar.  @themightychris  Let me know if you think a change like this would be warranted.

### DIFF
--- a/static-webapps/slate-portfolio-manager/src/views/AdvancedPortfolioManager/AdvancedPortfolioSidebar/AdvancedPortfolioSidebar.vue
+++ b/static-webapps/slate-portfolio-manager/src/views/AdvancedPortfolioManager/AdvancedPortfolioSidebar/AdvancedPortfolioSidebar.vue
@@ -139,7 +139,6 @@ export default {
       }));
     },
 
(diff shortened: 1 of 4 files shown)


themightychris (changes requested):
@BillClinton I think your suggestion makes a lot of sense, let's go ahead and do that to keep the code as understandable as possible

themightychris (approved):
LGTM

